PR #148, opened 2020-05-21
Major: Stateful validations

ealush:
| Q                | A
| ---------------- | ---
| Bug fix?         | ✖
| New feature?     | ✔
| Breaking change? | ✔
| Deprecations?    | ✔
| Documentation?   | ✔
| Tests added?     | ✔

Usage example: https://stackblitz.com/edit/vest-react-support-example-rdxuyh
Introducing stateful validations:

New interface: vest.create()

```js
const validation = data => {
  const validate = vest.create('NewUserForm', () => {
    test('username', 'Must be at least 3 chars', () => {
      enforce(data.username).longerThanOrEquals(3);
    });

    test('email', 'Is not a valid email address', () => {
      enforce(data.email)
        .isNotEmpty()
        .matches(/[^@]+@[^\.]+\..+/g);
    });
  });

  const res = validate();
}
```

# Understanding Vest's state

Vest is designed to help perform validations on user inputs. The nature of user inputs is that they are filled one by one by the user. In order to provide good user experience, the best approach is to validate fields as the user type, or when they leave the field.

The difficult part when validating upon user interaction is that we want to only validate the field that the user is currently interacting with, and not the rest of the form. This
This can be done with Vest's `only()` hook. That's where the state mechanism is becoming useful.

When you have skipped fields in your validation suite, vest will try to see if those skipped fields ran in the previous suite, and merge them into the currently running suite result - so the result object you get will include all the fields that your user interacted with.

Vest state is anchored to the suite name (defined in vest.create's first argument), which means the suite name must be unique.

## What Vest's state does

- _Skipped field merge_

As mentioned before - whenever you skip a field, vest will look for it in your previously ran suites and add it to the current result.

- _Lagging async `done` callback blocking_

In case you have an async test that didn't finish from the previous suite run - and you already ran another async test for the same field - vest will block the `done()` callbacks for that field from running for the previous suite result.


README.md
```diff
@@ -12,44 +12,48 @@
...
+It works great with user-input validation and with validating upon user interaction to provide the best possible user experience.
 
 The idea behind Vest is that your validations can be described as a 'spec' or a contract that reflects your form or feature structure. Your validations run in production, and they are framework agnostic - meaning Vest works well with React, Angular, Vue, or even without a framework at all.
 
 **Example code:**
 
 ```js
 // validation.js
-import { validate, test, enforce } from "vest";
+import vest, { validate, test, enforce } from 'vest';
```
Attrash-Islam, line 23: I think you don't need validate anymore. Right?

config/jest/jest.setup.js
```diff
@@ -1,26 +1,9 @@
...
 
-global.vestDistVersions = [];
-
-const isWatchMode = (process.argv || []).some(
-  (arg) => arg && arg.includes("--watch")
-);
-
-const VEST_DIST_FILES = glob.sync(
-  packagePath(PACKAGE_NAME_VEST, "dist", "*.js")
+global.isWatchMode = (process.argv || []).some(
```
Attrash-Islam, line 5: If I'm not wrong process.argv can't be nil (null or undefined). Isn't? Even I guess it won't be empty array either.

docs/README.md
```diff
@@ -19,37 +19,40 @@ The idea behind Vest is that your validations can be described as a 'spec' or a
 
 ```js
 // validation.js
-import { validate, test, enforce } from "vest";
+import vest, { validate, test, enforce } from 'vest';
```
Attrash-Islam, line 22: Same here regarding the validate import.

docs/exclusion.md
```diff
@@ -2,44 +2,65 @@
 
 When performing validations in real world-scenarios, you may need to only run tests of a single field in your suite, or skip certain tests according to some logic. That's why Vest includes `vest.skip()` and `vest.only()`.
 
-`vest.skip()` and `vest.only()` are functions that take a name of the test, or a list of names to either include or exclude fields from being validated. They should be called from the body of `validate` callback, and in order for them to take effect, they should be called before anything else.
+`vest.skip()` and `vest.only()` are functions that take a name of the test, or a list of names to either include or exclude fields from being validated. They should be called from the body of suite callback, and in order for them to take effect, they should be called before anything else.
 
-Fields that were skipped (or not included when `only` was used) will appear in the [result object](./result) in the `skipped` array.
```
Attrash-Islam, line 7: Putting the excluded in another `skipped` path won't make my life hard as consumer. Would be glad if that can be a boolean attribute where I can get the validation in both cases (skipped / Not skipped) the same way.

docs/exclusion.md
```diff
@@ -2,44 +2,65 @@
...
 When validating upon user interactions, you will usually want to only validate the input the user currently interacts with to prevent errors appearing in unrelated places. For this, you can use `vest.only()` with the name of the test currently being validated.
 
 In the example below, we're assuming the argument `fieldName` is being populated with the name of the field we want to test. If none is passed, the call to `only` will be ignored, and all tests will run as usual. This allows us to test each field at a time during the interaction, but test all on form submission.
 
 ```js
-import vest, { validate, enforce, test } from 'vest';
+import vest, { enforce, test } from 'vest';
 
-const v = (data, fieldName) => validate('New User', () => {
+const validate = (data, fieldName) => {
```
Attrash-Islam, line 20: It's a little bit harder for eyes to see const validate twice, would consider changing them to different things in order to make it easier to read.

packages/vest/dist/vest.js
```diff
@@ -1,21 +1,29 @@
...
-b.forEach(H)});return d.output},warn:function(){var b=l.useContext();b?b.currentTest?b.currentTest.warn():k("warn hook called outside of a test callback. It won't have an effect."):k("warn hook called outside of a running suite.")}})})
+'use strict';(function(v,r){"object"===typeof exports&&"undefined"!==typeof module?module.exports=r():"function"===typeof define&&define.amd?define(r):(v=v||self,v.vest=r())})(this,function(){function v(d,a){return a={exports:{}},d(a,a.exports),a.exports}function r(d){"@babel/helpers - typeof";r="function"===typeof Symbol&&"symbol"===typeof Symbol.iterator?function(a){return typeof a}:function(a){return a&&"function"===typeof Symbol&&a.constructor===Symbol&&a!==Symbol.prototype?"symbol":typeof a};
```
Attrash-Islam, line 1: Would consider getting this out and to put it in .gitignore

packages/vest/playground/index.js
```diff
@@ -0,0 +1,25 @@
+const vest = require('../dist/vest.min.js');
```
Attrash-Islam, line 1: I see why you keep them. But would consider other solutions. I even may think of installing the package as dependency to itself.

packages/vest/src/core/state/symbols.js
```diff
@@ -0,0 +1,3 @@
+export const SYMBOL_STATE = Symbol('vestState');
```
Attrash-Islam, line 1: What is the point of the string sent to the Symbol? I know that Symbol() !== Symbol() anyway. Are you using this string in some way or you're using the variables as unique values? (The second means that you may don't need the string values.

ealush: I use the descriptor for easier debugging.

packages/vest/playground/index.js
ealush: The playground is intended for live testing during development, having the package a dependency of itself means that I will get the latest production build, but not what I am currently working on.

packages/vest/dist/vest.js
ealush: This is intentional. The idea is that one can download vest as a script file to use on their site.
But maybe this really is not needed anymore in 2020  🤔

docs/exclusion.md
ealush: I am actually removing the skipped array in this version altogether. The `tests` object is sufficient in my opinion to know what's been tested and what's not.

packages/vest/playground/index.js
Attrash-Islam: I meant to `npm i .` not sure of the syntax but I know it works

PR #261, opened 2020-07-19
added: promisify utility

adife:
| New feature?     | ✔                                                                       |
| Breaking change? |✖                                                                        |
| Documentation?   | ✔                                                                    |
| Tests added?     | ✔                                                                      |

## `promisify()`

Promisify is a function that enables you to run your async validations as a [Javascript Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
This can be useful when running async validations on the server, or when you do not need the partial validation results. The Promise is resolved when all tests finish running.

### Usage

`promisify()` accepts a validation suite declaration, and returns a function that when called, returns a Promise.

```js
import vest from 'vest';
import promisify from 'vest/promisify';

const validate = promisify(
  vest.create('CreateNewUser', data => {
    test('email', 'The email already exists', () => doesEmailExist(data.email));
    test('username', 'The username already exists', () =>
      doesUsernameExist(data.username)
    );
  })
);

validate(data).then(res => {
  if (res.hasErrors('email')) {
    /* ... */
  }

  if (res.hasErrors('usename')) {
    /* ... */
  }
});
```


packages/vest/src/utilities/tests/promisify.test.js
```diff
@@ -0,0 +1,82 @@
...
+      expect(() => invalidValidateAsync()).toThrow();
+    });
+  });
+
+  describe('Return value', () => {
+    it('should be a function', () => {
+      expect(typeof promisify(jest.fn())).toBe('function');
+    });
+
+    it('should be a promise', () =>
```
raphaelboukara, line 27: i think u wanted to pass `done` here

adife: https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/no-test-callback.md

packages/vest/docs/utilities.md
```diff
@@ -70,3 +70,66 @@ vest.create('Checkout', () => {
...
+import promisify from 'vest/promisify';
+import validate from './validation.js';
+
+export default function Form() {
+  const validationResult = promisify(validate(data));
+
+  validationResult.then(result => {
+    result.hasErrors('email');
+  });
+}
```
ealush, line 134: I think this would make more sense as an example with `async/await`
```suggestion
// Form.js
import promisify from 'vest/promisify';
import validate from './validation.js';

const validateAsync = promisify(validate);

async function handler(data) {
  const result = await validateAsync(data)

  result.hasErrors('email')
}
```

PR #349, opened 2020-08-30
FEAT: [N4S] use Jest assertion return api

### LINKED ISSUE #285: [N4S] Add jest style enforce rule failures
At the moment, validation rules are simple functions that return a boolean. If the rule's return value is false, enforce will throw an Error.

Rule example:
https://github.com/ealush/vest/blob/69eacb388ad9fac8e403351df5ab5347d6298edf/packages/n4s/src/rules/greaterThan/index.js#L3-L5

And where it throws:
https://github.com/ealush/vest/blob/69eacb388ad9fac8e403351df5ab5347d6298edf/packages/n4s/src/enforce/runner/index.js#L14-L16

This means that the thrown error is always a generic one, not indicating the rule thrown, or the expected behavior of the rule.

Jest has a nicer interface for constructing matchers that also includes a custom failure message for that rule:
https://jestjs.io/docs/en/expect#expectextendmatchers

This can be very beneficial for Vest DEV_MODE that's being planned. A debug dump could log all the failed rules and their failure messages.

I would like to preserve the current behavior of Vest rules to prevent a breaking change, but also support the more verbose failure style of:
```js
{
	message: (ruleName, enforceValue, ...args) => ``, // someErrorMessage
    pass: true/false
}
```

vligas:
| New feature?     | ✔                                                                    |
| Fixed issues     | #285 |

Hi again 😄.

I tried to do this one without bother you. I encounter two ways of doing this.

1. Admiting the 2 types of return values: booleans and objects (this one is what I implemented)
2. Appending a message method to the original rules something somewhat similar to this
```js
import isNumeric from '../isNumeric';

function greaterThan(value, arg1) {
  return isNumeric(value) && isNumeric(arg1) && Number(value) > Number(arg1);
}

greaterThan.alias = 'gt';
greaterThan.message = (pass, value, arg1) => {
  ...
}
export default greaterThan;
```
PD: I also took the freedom to remove some dist folder that were beign tracked


packages/n4s/src/enforce/runner/index.js
```diff
@@ -11,8 +11,17 @@ function runner(rule, value, ...args) {
     return;
   }
 
-  if (rule(value, ...args) !== true) {
-    throw new Error(`[Enforce]: invalid ${typeof value} value`);
+  const ruleResult = rule(value, ...args);
+  // Handles boolean rules
+  if (typeof ruleResult === 'boolean' && ruleResult !== true) {
```
ealush: I believe this condition covers the same cases.
```suggestion
  if (ruleResult === false) {
```

packages/n4s/src/enforce/runner/index.js
```diff
@@ -11,8 +11,17 @@ function runner(rule, value, ...args) {
...
+    throw new Error(
+      `[Enforce]: invalid ${typeof value} value with rule ${rule.name}`
+    );
+  }
+
+  // Handles object rules { pass, message }
+  if (typeof ruleResult === 'object' && !ruleResult.pass) {
+    throw new Error(`[Enforce]: ${ruleResult.message}`);
   }
 }
```
ealush: This will fail on null return values because:

```js
typeof null // -> 'object'
```

Also, we did not cover the message function.


I would modify this approach a little to also support the message function. Feel free to make changes to it.

```suggestion
  if (ruleResult && ruleResult.pass === false) {
    const message = (typeof ruleResult.message === 'function'
    	? ruleResult.message() : ruleResult.message) || `invalid ${typeof value} value`;
  	throw new Error(`[Enforce]/${rule.name}: ${message}`);
  }
}
```

ealush (commented):
Hey @vligas, great effort! Thank you for taking the time.

There are a few things missing, though. I would really appreciate it if you could add them to your PR.

1. Support for function message value
2. Support for `ensure` as well as enforce https://github.com/ealush/vest/blob/latest/packages/n4s/src/ensure/runner/index.js

Also, for inspiration - here's how jest does it: https://github.com/facebook/jest/blob/master/packages/expect/src/index.ts#L264

Again, thank you for your time 🙏🏻

vligas:
Hi, i reviewed the jest code and I will be adding a `validateResultShape` and a `transformResultShape` functions so in first place we could parse all supported formats to a final one with defaults values.
```js
transformResultShape(true) // { pass: true, message: '[Enforce] Generic message'}
transformResultShape({ pass: true, message: 'Custom Message'}) // { pass: true, message: '[Enforce] Custom Message'}
transformResultShape({ pass: true }) // { pass: true, message: '[Enforce] Generic message' }
```
that way to deprecate/add a valid rule result, you only need to change those two functions and the rest of the lib could be only aware of the verbose one.

In the other hand what should `transformResultShape` in case that the rule returned an invalid format? pass silently? or maybe return a failing test like `{ pass: false, message: '[Enforce] Wrong rule result format' }`

what do you think @ealush?

ealush:
Hey, I like this approach. Was aiming towards something like this so you can reuse this function both in enforce and in ensure.

Now, transformResultShape should probably accept an additional, first, argument, such as `interfaceName`, and it could be either `Enforce` or `Ensure` - so you call it like this:

```js
transformResultShape('Enforce', ruleResult);
```

It might be worth looking into binding the first argument by default so the runner file would look like that:

```js
// enforce/runner/index.js
import transformResultShape from '../transformResultShape';

const transformResult = transformResultShape.bind(null, 'Enforce');
```
Not sure that it would make it nicer, though, because we only need it once per file. What do you think?

---
Regarding the malformed structure - I am following the [Jest validation](https://github.com/facebook/jest/blob/master/packages/expect/src/index.ts#L370) in a way that makes sense to use in Vest:

1. Any value other than explicit true or the result object - we throw a bad rule error. Ideally we should find a way to notify the consumer of this. I think we can add it to the development build in the future.

2. An object with `pass: false` without a message - fails with the generic rule message
3. An object without a boolean `pass` - bad rule error.

@vligas What do you think?

vligas:
Sure, I didn't took into account the interface name, maybe in the definition of the helper function we could add those binds maybe change the function name to `transformResult` and exporting `transformResultEnforce` and `transformResultEnsure`.

And I'm good with the rules that you mention I think they made logic, about how to let the user know if their rule is malformed I think the best woul be to add something like a console.error and letting return a failure (like a failing test) with a custom error that says HEY YOU ARE DOING THIS WRONG (not literally 😆). also we could separate these 2 scenarios (rule didn't pass && rule with malformed return) into separate error classes.

Either way I will be working in the helpers and letting the space open to implement the throw, let me know if you have other ideas on the subject 👍.

package.json
```diff
@@ -46,7 +46,7 @@
   "scripts": {
     "build": "node ./scripts/buildAll",
     "prettier-watch": "onchange '**/*.js' '**/*.json' -- prettier --write {{changed}}",
-    "test": "jest --projects ./packages/*/",
+    "test": "jest --projects ./packages/n4s",
```
ealush: I think you left this here by mistake 😅

packages/n4s/src/lib/transformResult/index.js
```diff
@@ -0,0 +1,62 @@
...
+ * @returns {boolean} result.pass indicates if the test passes or not
+ */
+export function transformResult(interfaceName, result, { rule, value }) {
+  const defaultResult = getDefaultResult(interfaceName, value, rule);
+
+  if (!isValidResult(result)) {
+    // #TODO handle invalid result
+    throw new Error(
+      `[${interfaceName}]/${rule.name} wrong return value for the rule please check that the return is valid`
+    );
```
ealush: Maybe it is best here to use the `throwError` module: https://github.com/ealush/vest/blob/latest/packages/n4s/src/lib/throwError/index.js

There are two reasons behind it:
1. It can be easily mocked in tests
2. We can later easily integrate it with the `__DEV__` mode and the development bundle.

Also, looking at the code in throwError just realized that we might not need the interfaceName after all. I think I mislead you here. We have a global that gets added during build. It's called `LIBRARY_NAME`.

If you don't mind changing this part back (basically, remove the interfaceName parts), it would be awesome. Otherwise, I'll change it back myself when merging your code.

packages/n4s/src/lib/transformResult/index.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {function} options.rule
+ * @param {*} options.value
+ * @returns {Object} result
+ * @returns {string} result.message
+ * @returns {boolean} result.pass indicates if the test passes or not
+ */
+export function transformResult(interfaceName, result, { rule, value }) {
+  const defaultResult = getDefaultResult(interfaceName, value, rule);
+
+  if (!isValidResult(result)) {
```
ealush: You're triple-negating (`!!!`) your value here:

```!isValidResult(result)```
```
export function isValidResult(result) {
  return !!(
  // ..
```


---
You could save all this flexing with the pattern they use at jest: https://github.com/facebook/jest/blob/master/packages/expect/src/index.ts#L268

They call `validateResult`, which is responsible for immediately throwing - which is exactly what you want to do.

A side bonus of that is that you will be able to test that we throw in the validation stage - which is harder to test at the moment.

File: packages/n4s/src/lib/transformResult/index.js [hunk 6]
```diff
@@ -0,0 +1,62 @@
...
+    // #TODO handle invalid result
```
ealush: Is this TODO still needed? I think you already do handle the invalid result.

packages/n4s/src/lib/transformResult/index.js
```diff
@@ -0,0 +1,62 @@
...
+        rule,
+        typeof result.message === 'function' ? result.message() : result.message
+      );
+    }
+    return { ...defaultResult, ...formattedResult };
+  }
+}
+
+export const transformResultEnforce = transformResult.bind(null, 'Enforce');
+export const transformResultEnsure = transformResult.bind(null, 'Ensure');
```
ealush: So, considering we have `LIBRARY_NAME`, we don't really need these two. Again, sorry about that. 😅

ealush (approved):
Great job!
I am approving this PR even though I have a few more comments. I think we are close enough, and in general I think that the progress in this PR is great.
I will merge once we're ready with these changes as well.

Also, I think I mislead you with the interfaceName stuff. Sorry about that. Let me know if you prefer that I fix it.

package.json
vligas: oh yes, I'am working on W10 and I have some problems running the tests for all the packages, my bad 🤦

packages/n4s/src/lib/transformResult/index.js
vligas: Ohh yes i did but i was waiting also to know what where you thoughts on that part, and as you say I'll be changing that part for the throwError func. just a reminder 👍

packages/n4s/src/lib/transformResult/index.js
vligas: Ok, that double negation was because `isValidResult(null)` was returning `null`. I was not intending to triple negate, just a shortcut to ensure that it was returning a boolean.

I see the benefits of the `validateResult`. I'll be changing that

packages/n4s/src/lib/transformResult/index.js
vligas: Sure, I can take it back 🙆‍♂️

packages/n4s/src/lib/transformResult/index.js
vligas: Hey no problem, thanks for the reviews 👌

vligas:
Well, I think is done, I had some hard time mocking the `throwError` function to check if it was beign called by the `validateResult`. If you could take a look at adding those excpects in the `validateResult` tests when you merge please ><

packages/n4s/src/lib/transformResult/spec.js
```diff
@@ -0,0 +1,130 @@
...
+  goodObjectMessageFunctionRule,
+} from '../../testUtils/rules';
+import {
+  transformResult,
+  getDefaultResult,
+  formatResultMessage,
+  validateResult,
+} from '.';
+
+describe('Tests `isValidResult` helper', () => {
```
ealush:
```suggestion
describe('Tests `validateResult` helper', () => {
```

packages/n4s/src/lib/transformResult/spec.js
```diff
@@ -0,0 +1,130 @@
...
+        transformResult(result, { rule: nullRule, value: bool })
+      ).toThrow(Error);
+    });
+  });
+});
+
+describe("Test transform result's message", () => {
+  [goodObjectMessageRule, goodObjectMessageRule, goodObjectRule].forEach(
+    rule => {
+      it('Should containt de name of the lib', () => {
```
ealush:
```suggestion
      it('Should contain the library name', () => {
```

packages/n4s/src/lib/transformResult/spec.js
```diff
@@ -0,0 +1,130 @@
...
+
+  it("Should return false if it is a verbose result and doesn't have pass property", () => {
+    const missingPass = () => ({ message: 'missing pass' });
+    expect(() => validateResult(missingPass())).toThrow(Error);
+  });
+});
+
+describe('Test transform result', () => {
+  // to make sure we don't end up leaving some hardcoded value behind
+  [true, false].forEach(bool => {
```
ealush: Looks like `describe.each([true, false])` suits your needs better.

https://jestjs.io/docs/en/api#describeeachtablename-fn-timeout

ealush (commented):
😍
Yep, everything looks really good.

Some strings should be changed in the tests, but other than that, it looks amazing.

ealush:
YES!
Thank you. This feature can over time really improve Vest for consumers. I appreciate you taking the time to work on it.

PR #351, opened 2020-08-30
patch: Create a "suite" core folder

### LINKED ISSUE #312: Restructure state directory
The idea is to move all the modules that directly relate to a suite and put them in the same directory: `packages/vest/src/core/suite`.

- [ ] create core/suite directory
  - [ ] move createSuite to suite/create
  - [ ] move validate to suite/validate
  - [ ] move registerSuite to suite/register
  - [ ] move reset to suite/reset
  - [ ] move remove to suite/remove
  - [ ] move getSuiteState to suite/getState
  - [ ] move patch to suite/patch
  - [ ] move cleanupCompletedSuite to suite/cleanupCompleted
  - [ ] move getSuite to suite/getSuite (maybe just get? need to see how it feels to actually use it)
  - [ ] move setSuite to suite/setSuite (maybe just set? need to see how it feels to actually use it)
  - [ ] Move hasRemainingTests to suite/hasRemainingTests

NorbertLuszkiewicz:
| New feature?     | ✖                                                                        |
| Tests added?     | ✖                                                                        |
| Fixed issues     | https://github.com/ealush/vest/issues/312 |

I Moved all the modules that directly relate to a suite,
I unified the names, e.g. from getSuiteState to getState


ealush (changes requested):
Hey @NorbertLuszkiewicz, thank you for taking the time 🙏🏻
Looks like some of the tests are failing and the build does not finish successfully.

image

Can you please run another pass at the change to make sure that all the files are reachable?

Tip: Try running `yarn lint`, it will show you all the files that it is unable to resolve.

image

NorbertLuszkiewicz:
@ealush Can you take a look at my changes? ? :D

packages/vest/src/core/state/cleanupCompletedSuite/index.js
```diff
@@ -1,12 +1,12 @@
-import getSuiteState from '../getSuiteState';
```
ealush, line 1: I believe that this file (and its directory) should go.

packages/vest/src/core/suite/getState/index.js
```diff
@@ -1,4 +1,4 @@
-import { getSuite } from '..';
+import { getSuite } from '../../state';
 
 /**
  * Retrieves most recent suite state.
```
ealush, line 4: The function in this file should probably be renamed.

packages/vest/src/core/suite/getState/spec.js
```diff
@@ -1,6 +1,6 @@
-import * as state from '..';
 import runRegisterSuite from '../../../../testUtils/runRegisterSuite';
-import getSuiteState from '.';
+import * as state from '../../state';
+import getState from '.';
 
 const suiteName = 'suite_1';
 const suiteId = 'suiteId_1';
```
ealush, line 6: The top describe block still references `getSuiteState`

ealush (commented):
This is much better! Thank you.

I did notice some naming inconsistencies. In some places you kept the old function names, and in other you did follow up on the rename. I commented on a few, but there are more.
The way I would go about it - go over to the renamed files (or directories) - and make sure they declare the function they are actually supposed to. Same for the spec files.

@NorbertLuszkiewicz Can you please run another pass on these changes to make the naming consistent?

ealush:
These changes make me wonder if we should also add a `suite/index.js` module that would then re-export all these modules so when consuming these files we would prevent confusion:

```js
import suite from '../suite';

// .. some code
suite.getState()
```

But that's for a different change.

packages/vest/src/core/suite/getState/index.js
NorbertLuszkiewicz: but I don't know what name is correct get is busy in several files

packages/vest/testUtils/resetState/index.js
```diff
@@ -1,19 +1,19 @@
...
-      suiteId,
-      name: suiteName,
+  if (Name) {
+    runRegister({
+      Id,
+      name: Name,
       operationMode: OPERATION_MODE_STATEFUL,
     });
-    return getSuiteState(suiteId);
+    return getState(Id);
```
ealush: Looks like the tests are failing due to this change. The runRegister function ends up here: https://github.com/ealush/vest/blob/4f246971f7b280e2bf02b3b0d2fe45f56547affd/packages/vest/src/core/state/index.js#L27

This function expects suiteId and not `Id`

You can fix nearly all of the failed tests by reverting these names back (suiteId, suiteName). I am assuming this was just a bad copy/paste - I see you even kept the capitalized I and N.

ealush (commented):
This looks much better!
The tests are failing though. I traced most of them to the resetState testUtil.

There may be a few more failing tests due to other changes - but the solution should be similar.

ealush (approved):
Hey @NorbertLuszkiewicz, well done!


I noticed two tiny things, so I went ahead and fixed those, but I am really pleased with the result. Thank you.

In the next step I will create an index module to the suite folder that will make use of your changes.

I appreciate all your help 👏🏻

PR #357, opened 2020-08-31
chore: Add __DEV__ to true in dev build mode

genuineswe:
| Tests added?     | ✖                                                                        |
| Fixed issues     | as in issue 310, providing more runtime metrics for development build |

Only add NODE_ENV and __DEV__ as constant and attribute.

### DIFF
--- a/config/builds/vest.js
+++ b/config/builds/vest.js
@@ -23,6 +23,8 @@ const FORMAT_ES = 'es';
 const FORMAT_CJS = 'cjs';
 const DIR_ESM = 'esm';
 
+const { NODE_ENV } = process.env;
+
 const DIST_PATH = path.join(PACKAGE_PATH, DIR_NAME_DIST);
 
 const plugins = ({ name, format }) => {
@@ -56,6 +58,7 @@ const plugins = ({ name, format }) => {
     replace({
       VEST_VERSION: JSON.stringify(version),
       LIBRARY_NAME: JSON.stringify(PACKAGE_VEST),
+      __DEV__: NODE_ENV === "development" ? true : false
     }),
   ];
 


ealush (commented):
Hey @gaweki, thank you for taking the time to improve Vest 🙏🏻

Maybe I should have been a little more descriptive when opening the issue, because as it is now it might lead to unexpected results.

By itself the change looks fine, but if you look below - I build all bundles in the same run, so we cannot reliably use NODE_ENV here.

https://github.com/ealush/vest/pull/357/files#diff-577067326a70544e1915963cc4360576L102-L117

We might change this in the future.

Instead, I set the build `name` to development, you can use this variable name in your condition:

```
name === ENV_DEVELOPMENT
```

Now, when using this variable in the built version it should be fine, but Jest has no knowledge of it, so we need to add it as a global variable to our jest setup right next to the other declarations:

https://github.com/ealush/vest/blob/latest/packages/vest/config/jest/jest.setup.js#L6

And to prevent eslint from shouting, it should be added as a global as well:
https://github.com/ealush/vest/blob/latest/.eslintrc.js#L19

Again, thank you. I appreciate your help!

genuineswe:
Hello @ealush , i add ENV_DEVELOPMENT : true in [here](https://github.com/ealush/vest/blob/latest/.eslintrc.js#L19) and add global.ENV_DEVELOPMENT = true in [here](https://github.com/ealush/vest/blob/latest/packages/vest/config/jest/jest.setup.js#L6) ?

then in [line 59](https://github.com/ealush/vest/blob/latest/config/builds/vest.js#L56-L59), i change to __DEV__: name === ENV_DEVELOPMENT. but where the "name" constant imported from?

ealush:
1. Yes 🙂
2. Yes 🙂
3. `name` is one of the arguments of the `plugins` function. If you take a look a couple of lines above, it is actually already in similar use:
https://github.com/ealush/vest/blob/latest/config/builds/vest.js#L38-L40

What I would do here is probably create an isDevelopment constant inside this function and use that constant in both places.

I hope that helps!

genuineswe:
I'm sorry, @ealush . i thought PLUGINS array is outside of plugins function. Is it correct i make new PR?

ealush:
Continued in https://github.com/ealush/vest/pull/366

PR #419, opened 2020-09-30
feat: add n4s rule isBetween and isNotBetween

### LINKED ISSUE #396: [N4S] RULE add isBetween/isInRange
Add a new enforce rule:

```js
enforce(5).isBetween(0, 5) // ✅
enforce(5).isBetween(0, 10) // ✅
enforce(5).isBetween(0, 2) // 🚨
```

syncush:
| Documentation?   | ✔                                                                        |
| Fixed issues     | |

### DIFF
--- a/packages/n4s/src/rules/isBetween/index.js
+++ b/packages/n4s/src/rules/isBetween/index.js
@@ -1,0 +1,10 @@
+import gte from '../greaterThanOrEquals';
+import lte from'../lessThanOrEquals';
+
+function isBetween(value, min, max) {
+    return gte(value, min) && lte(value, max);
+}
+
+isBetween.negativeForm = 'isNotBetween';
+
+export default isBetween;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9386 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@ampproject/remapping": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-0.2.0.tgz",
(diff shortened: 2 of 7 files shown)


packages/n4s/src/rules/isBetween/index.js
```diff
@@ -0,0 +1,11 @@
+import gte from '../greaterThanOrEquals';
+import lte from'../lessThanOrEquals';
+
+function isBetween(value, arg1, arg2) {
+    const isCorrectType = typeof value === 'number' && typeof arg1 === 'number' && typeof arg2 === 'number';
+    return isCorrectType && (arg1 <= arg2 ? gte(value, arg1) && lte(value, arg2) : gte(value, arg2) && lte (value, arg1));
+}
```
ealush: Wouldn't this be sufficient?
```suggestion
function isBetween(value, arg1, arg2) {
  return gte(value, arg1) && lte(value, arg2);
}
```

We don't really are about the `typeof`, the value or the args could be numeric strings (`"10"`), and both lte and gte already do the parsing.

And regarding the second part, I understand it is intended to allow flipping the numbers, but just for brevity, I think we could safely assume `isBetween(value, min, max)`. I believe 90% of the consumer will assume that immediately, and the other 10% could rely on the docs to tell them how to construct it.

ealush (commented):
Good job implementing this feature. I think we could reduce it a little though if we ignore a specific use case.

packages/n4s/src/rules/isBetween/index.js
syncush: I agree that the `typeof` is redundant,  I did not assume a `min, max` I can remove the second use case. will do!

ealush (approved):

PR #433, opened 2020-10-13
next : isNegative & isPositive feature implementation

### LINKED ISSUE #401: [N4S] Rule isNegative | isPositive
Add a new enforce rule:

```js
enforce(-5).isNegative() // ✅
enforce(5).isNegative() // 🚨
```

```js
enforce(5).isPositive() // ✅
enforce(-5).isPositive() // 🚨
```

ganeshpatil0101:
| Documentation?   | ✔                                                                        |
| Fixed issues     | NA |

isNegative & isPositive feature implementation.


packages/n4s/src/rules/isNegative/index.js
```diff
@@ -0,0 +1,8 @@
+function isNegative(value, arg1) {
+  if (!isNaN(Number(value))) {
```
ealush: I think you could use here [`isNumeric`](https://github.com/ealush/vest/blob/latest/packages/n4s/src/rules/isNumeric/index.js). There's nothing inherently wrong with the way you're doing it, but this will maintain consistency with the rest of the number centric rules.

packages/n4s/src/rules/isPositive/index.js
```diff
@@ -0,0 +1,8 @@
+function isPositive(value, arg1) {
```
ealush: I actually think we can remove the `isPositive` rule, and instead, just use the `negativeForm` of `isNegative`.

Removing this file and using the other notation should also fix your tests.

packages/n4s/src/rules/isNegative/index.js
```diff
@@ -0,0 +1,8 @@
...
+    return Number(value) < 0;
+  }
+  return false;
+}
+
+export default isNegative;
```
ealush: Adding this will allow you to completely remove the isPositive file and its tests.

```suggestion
}

isNegative.negativeForm = 'isPositive';

export default isNegative;
```

ealush (commented):
Nice Job @ganeshpatil0101! Thank you 🙏🏻

Please note that apart from these small comments, there are a couple of lint errors. Those should be pretty easy to fix as well.

packages/n4s/src/rules/isNegative/index.js
ganeshpatil0101: used isNumeric function.

packages/n4s/src/rules/isPositive/index.js
ganeshpatil0101: removed isPositive rules. and added negativeForm inside isNegative rule and exposed isPositive.

ganeshpatil0101:
@ealush Thanks for the review. I made changes per review comments. Please check. regarding lint-errors , which exactly lint-errors?

packages/n4s/src/rules/isNegative/index.js
```diff
@@ -0,0 +1,10 @@
+import isNumeric from '../isNumeric';
+
+function isNegative(value, arg1) {
```
ealush: You're not using `arg1`.
```suggestion
function isNegative(value) {
```

ganeshpatil0101: Done.

ganeshpatil0101:
@ealush I am getting below error while running demo example.
Loading module from “http://127.0.0.1:8080/style.scss” was blocked because of a disallowed MIME type (“text/x-scss”).
do you have any idea why this getting?

ganeshpatil0101:
for more reference here is the screenshot https://ibb.co/6X4rs5T

ealush:
@ganeshpatil0101 Hm. This seems to be some kind of a problem with the sandbox and not with vest itself. Is anything actually not working as expected?

ealush (approved):
This is great! Thank you @ganeshpatil0101 for helping improve Vest. This will be added as part of Vest's next version in a couple of weeks 🙏🏻

ganeshpatil0101:
@ealush Thank You so much. I learnt many new things working with you and this enhancement.

Issue #269, opened 2020-07-25
Add enforce.any() rule


ealush:
At the moment, when chained, all rules have an `AND` relationship.

```js
enforce(4).isNumeric().isNumber().largerThan(2);
/*
enforce that 4
is numeric AND
is of type number AND
is larger than 2
*/
```

It can be useful to also have OR relationships between enforcement rules such as

**OR RELATIONSHIP**
At least one rule must be truthy
```js
enforce(value).any(
  enforce.isEmpty(),
  enforce.greaterThan(10)
);
```

This used to be difficult to achieve, but with the recent change https://github.com/ealush/vest/pull/479 that allows for lazy evaluated enforcements (`enforce.isArray()([])`) it should be pretty straightforward.

## Instructions:
To get this working, you need to add a new rule, you can call it `any` or `anyOf`, and have it take multiple functions. It should pass each of these functions the enforced value.

What we have at the moment is this:
```js
enforce.isNumber() // -> returns a function that takes a value
enforce.isNumber()(5) // -> returns true
enforce.isNumber()('not_a_number') // -> returns false
```

The assumption is that the consumer will do this:
```js
enforce(5).any(
  enforce.isNumber(),
  enforce.isUndefined()
);
```

and the `any` rule, will call each of the functions with 5.

See [Shape](https://github.com/ealush/vest/blob/latest/packages/n4s/src/enforce/compounds/shape.js) for reference.

ealush:
Alternative option, which would require some changes to enforce:

This requires enforce itself to also be some sort of a proxy and not just return one.

`enforce(value).anyOf(enforce.longerThan(5), enforce.isEmpty())`

ealush:
Rule updated with instructions

ealush:
closed by #493

PR #493, opened 2020-11-11
implement any rule #269

kapalex:
| Q                | A   |
| ---------------- | --- |
| New feature?     | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✔ |

Implementation matches the description in the issue. Decided on the name anyOf() as any() exists in vest in some form and causes issues with unique names.

### DIFF
--- a/packages/n4s/src/enforce/compounds/__tests__/anyOf.test.js
+++ b/packages/n4s/src/enforce/compounds/__tests__/anyOf.test.js
@@ -22,6 +22,15 @@ describe('AnyOf validation', () => {
         )
       ).toBe(true);
     });
+    it('Should succeed when rule chaining', () => {
+      expect(
+        anyOf(
+          [1, 2, 3],
+          enforce.isArray().isNotEmpty().longerThan(2),
+          enforce.isUndefined()
+        )
+      ).toBe(true);
+    });
     it('Should fail with no rules', () => {
       expect(
         anyOf(5)
--- a/packages/n4s/src/enforce/compounds/__tests__/anyOf.test.js
+++ b/packages/n4s/src/enforce/compounds/__tests__/anyOf.test.js
@@ -1,0 +1,50 @@
+import enforce from 'enforce';
+import anyOf from 'anyOf';
+
+describe('AnyOf validation', () => {
+  describe('Base behavior', () => {
+    it('Should fail when all rules fail', () => {
+      expect(
+        anyOf(
+          'test',
+          enforce.isNumber(),
+          enforce.isUndefined()
+        )
+      ).toBe(false);
+    });
+    it('Should succeed when atleast one rule applies', () => {
+      expect(
+        anyOf(
+          5,
+          enforce.isString(),
+          enforce.isNumber(),
+          enforce.isUndefined()
+        )
+      ).toBe(true);
+    });
+    it('Should fail with no rules', () => {
+      expect(
+        anyOf(5)
+      ).toBe(false);
+    });
+  });
+
+  describe('As part of enforce', () => {
+    it('Should validate anyof the rules correctly', () => {
+      expect(
+        enforce(77).anyOf(
+          enforce.isString(),
+          enforce.isNumber(),
+          enforce.isUndefined()
+        )
+      );
+
+      expect(() =>
+        enforce({ test: 4 }).anyOf(
+          enforce.isNumber(),
+          enforce.isUndefined()
+        )
+      ).toThrow();
+    });
+  });
+});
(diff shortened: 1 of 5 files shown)


packages/n4s/src/enforce/compounds/anyOf.js
```diff
@@ -0,0 +1,3 @@
+export default function anyOf(value, ...rules) {
+  return Boolean(!rules) || rules.some(fn => fn(value));
```
ealush: Not sure what you were trying to achieve with the first condition - rules will always be an array, hence, always truthy, so that condition will always fail. Maybe you meant to check `rules.length`? Anyway, in our case, since you went with `fail` for no rules, your current implementation should just work.
```suggestion
  return rules.some(fn => fn(value));
```

packages/n4s/src/enforce/compounds/__tests__/anyOf.test.js
```diff
@@ -0,0 +1,50 @@
...
+          enforce.isString(),
+          enforce.isNumber(),
+          enforce.isUndefined()
+        )
+      ).toBe(true);
+    });
+    it('Should fail with no rules', () => {
+      expect(
+        anyOf(5)
+      ).toBe(false);
```
ealush: Ok, let's go with that 👍

ealush (commented):
Great job @kapalex! this is looking really good. I am missing the typings, though. Maybe you added it on the untracked (built) file?
The types file is this: `./packages/vest/src/typings/vest.d.ts`

ealush (commented):
Great job @kapalex! this is looking really good. I am missing the typings, though. Maybe you added it on the untracked (built) file?
The types file is this: `./packages/vest/src/typings/vest.d.ts`

packages/n4s/src/enforce/compounds/__tests__/anyOf.test.js
kapalex: Well I was sure I actually went the way we agreed, and it was indeed supposed to be true and rules.length above.
I'll take care of it and the typing issue first thing tomorrow 😄

kapalex:
Everything should now be in order.  Fixed some minor issues the ci pointed out as well.
Not entirely sure about the typings, would appreciate your feedback

ealush:
As you'll probably notice, if you rebase from `latest` branch now, your tests will stop working. That's not due to an error on your end, but due to an api enhancement I made.

In short, lazy enforcement rules now support chaining, so the consumer can do something like this:

```js
enforce.isArray().isNotEmpty(1).shorterThan(5)
```

The way it internally works is like this: each rule returns all of the other rules, and a `.test()` function. Calling the `test` function will run all the chained rules at once, and pass its provided value.

You don't need to change your tests, but I will add instructions across your PR on how to fix those errors.

packages/n4s/src/enforce/compounds/anyOf.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * @param {*} value   Value to be test against rules
+ * @param {Function[]} rules    Rules to validate the value with
+ */
+export default function anyOf(value, ...rules) {
+  return !(rules.length) || rules.some(fn => fn(value));
```
ealush: As I described in my comment, the `fn` might actually be a group of functions. It can be a chain of `enforce.isNumber().greaterThan(5)`... So just calling the function as is won't work. You're going to need to call the `test` function. I added a helper function that will help with that: `runLazyRules`. This is how you can fix that:
```suggestion
  return !rules.length || rules.some(ruleGroup => runLazyRules(ruleGroup, value));
```

packages/n4s/src/enforce/compounds/anyOf.js
```diff
@@ -0,0 +1,7 @@
+/**
```
ealush, line 3:
```suggestion
import runLazyRules from 'runLazyRules';

/**
```

packages/vest/src/typings/vest.d.ts
```diff
@@ -258,6 +259,7 @@ type TEnforceLazy = {
     [key: string]: LazyEnforceValue | LazyEnforceValue[];
   }) => LazyEnforceValue;
   optional: <T>(...rules: LazyEnforceValue[]) => LazyEnforceValue;
+  anyOf: <T>(...rules: LazyEnforceValue[]) => LazyEnforceValue;
```
ealush: You're also going to get a merge conflict here. I renamed this key.
```suggestion
anyOf: <T>(...rules: TEnforceLazy[]) => TEnforceLazy;
```

ealush (commented):
Real exciting. I made quite a few changes across the repo, so I added some comments addressing the issues you might experience merging back from latest branch. Other than that, everything looks great!

kapalex:
Everything fixed to fit new changes, and added a test just to make sure it works correctly with chaining 😃

ealush (approved):
Niiice! Two in one day 🤩

Thank you @kapalex for helping to improve Vest 
🙏🏻

Issue #488, opened 2020-11-05
Add enforce.arrayOf() rule


ealush:
This rule can be quite useful when a future schema validation rule will be added to enforce.

The general idea is quite similar to the [enforce.any](https://github.com/ealush/vest/issues/269) rule:

```js
enforce(value).isArrayOf(
  enforce.isString(),
  enforce.isNumber(),
 enforce.isArrayOf(...) // Yeah, why not? A recursive call should be pretty simple here
)
```

The idea is that at least one needs to be truthy in order for the condition to be met - alternatively, an empty array should probably be valid as well unless explicitly mentioned:

```js
enforce(value).isArrayOf(
  enforce.isString()
).isNotEmpty()
```

See [Shape](https://github.com/ealush/vest/blob/latest/packages/n4s/src/enforce/compounds/shape.js) for reference.

Moses3301:
#499 implement isArrayOf rule

ealush:
Closed by #499

PR #499, opened 2020-11-13
added: isArrayOf rule [N4S] #488

Moses3301:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✖ |
| Tests added?     | ✔ |
| Types added?     | ✖ |
| Related issues   |     |

Add isArrayOf  rule


packages/n4s/src/rules/rules.js
```diff
@@ -27,6 +27,7 @@ import { numberEquals, numberNotEquals } from 'numberEquals';
 import { shorterThan } from 'shorterThan';
 import { shorterThanOrEquals } from 'shorterThanOrEquals';
 import { startsWith, doesNotStartWith } from 'startsWith';
+import { isArrayOf } from 'isArrayOf';
```
ealush: I think this file export would be more suitable with `compounds` rather than rules, so instead of importing and exporting it here, it would probably make more sense to do it here:

https://github.com/ealush/vest/blob/latest/packages/n4s/src/enforce/compounds/compounds.js

packages/n4s/src/rules/__tests__/isArrayOf.test.js
```diff
@@ -0,0 +1,40 @@
+import { isArrayOf } from 'isArrayOf';
+import enforse from 'enforce'
```
ealush: typo 🙈
```suggestion
import enforce from 'enforce'
```

package.json
```diff
@@ -6,7 +6,7 @@
     "prettier-watch": "onchange '**/*.js' '**/*.json' -- prettier --write {{changed}}",
     "dev": "onchange -i './packages/**/src/**/*.js' -- yarn genJSConfig",
     "genJSConfig": "node ./scripts/genJsconfig",
-    "test": "jest --projects ./packages/*",
+    "test": "jest --projects ./packages/*",
```
ealush:
```suggestion
    "test": "jest --projects ./packages/*",
```

packages/n4s/src/rules/isArrayOf.js
```diff
@@ -0,0 +1,4 @@
+export function isArrayOf(value, ...rules) {
+    if (value.some(element => !rules.some(fn => fn(element)))) {return false};
+    return true;
+  };
```
ealush: There are a couple of things I would consider changing in this function.

1. value might be `undefined`, and in some cases that would be ok. I would add an initial isArray safeguard at the beginning, and if value is not an array, return false immediately. At the moment, this function would throw.

2. I think you could skip the negation (!) on `!rules.some` if you switch `value.some` with `value.every`.

3. The inner function `fn => fn(element)` wouldn't work anymore due to a [recent change](https://github.com/ealush/vest/commit/40591805dcf5b00bfc005895eb0303491c66958a#diff-4e3cd2887d1ed9da69b47db120ccee942a92e9837efc21d70fc76409374b61afR80) I made to the lazy rules. They now support chaining as well, and to run them you need to call `.test`. You could use the `runLazyRules(fn, element)`

```suggestion
import runLazyRules from 'runLazyRules';

export function isArrayOf(value, ...rules) {
  return !value.every(element => rules.some(ruleChain => runLazyRules(ruleChain, element)))
};
```

packages/n4s/src/rules/__tests__/isArrayOf.test.js
```diff
@@ -0,0 +1,40 @@
...
+  
+    it('Should return false if one element or more fails all rules', () => {
+      expect(
+        isArrayOf([3,4,5,['s','i','x'],7],
+        enforse.greaterThan(2),
+        enforse.isArrayOf(enforse.isNumber()))
+        ).toBe(false);
+    });
+  });
+
```
ealush: I would also add one integration test that makes sure everything works together with enforce

Here's an example, should probably be more thorough.

```js
describe('as part of enforce', () => {
  it('should return silently when valid', () => {
    enforce([1,2,'3']).isArrayOf(enforce.isNumber())
    enforce([1,2,'3']).isArrayOf(enforce.isString())
    enforce([1,2,'3']).isArrayOf(enforce.isNumeric(), enforce.lessThan(5).greaterThan(0))
  });

  it('should throw an exception when invalid', () => {
    expect(() => enforce([1,2,'3']).isArrayOf(enforce.isNull())).toThrow()
    expect(() => enforce([1,2,'3']).isArrayOf(enforce.isNumber(), enforce.greaterThan(5))).toThrow()
  });
})
```

packages/n4s/src/rules/__tests__/isArrayOf.test.js
```diff
@@ -0,0 +1,40 @@
...
+    expect(isArrayOf([3,4,5,['s','i','x'],7], 
+    enforse.greaterThan(2), 
+    enforse.isString())
+    ).toBe(false);
+  });
+
+
+  describe('Tests for recursive call', ()=>{
+
+    it('Should return true if all elements are ture for one or more rules', () => {
```
ealush: typo 🙈
```suggestion
    it('Should return true if all elements are true for one or more rules', () => {
```


also here: 
https://github.com/ealush/vest/pull/499/files#diff-e2dfd7aa604cef61b1c17ca8cf0a9e4fc5bbe176d63350fbf4a01a6d95f68219R6

ealush (commented):
Great job @Moses3301, thank you for helping improve Vest! 🙏🏻
Everything looks great, only some minor changes.

First, I made quite a big change around lazy enforced rules, so rebasing back from latest branch would cause your tests to fail. This is easily fixable by applying the changes I described in your code.

Also, you are missing typescript types. It should be fairly simple to implement. Basically, all you need to do is duplicate this line: https://github.com/ealush/vest/blob/latest/packages/vest/src/typings/vest.d.ts#L260 and replace `optional` with `isArrayOf`.

Last thing, I see some lint errors in the build. You should run `yarn lint`. It should let you know what's wrong, and offer to auto-fix most of it.

Thank you again for the awesome work! I appreciate it. 🙏🏻

packages/n4s/src/rules/__tests__/isArrayOf.test.js
Moses3301: Thank you for the replay,
I'm working on it, and I have a question
in the `'as part of enforce' -> 'should return silently when valid'` test is 
```
enforce([1,2,'3']).isArrayOf(enforce.isNumber())
enforce([1,2,'3']).isArrayOf(enforce.isString())
```
need to return silently? '3' isn't  a number and '1','2' isn't a string.
did you meant to write
`enforce([1,2,'3']).isArrayOf(enforce.isNumber(), enforce.isString())`?

ealush: Ahh, right. You are correct. I just wrote it as an example. Should be more similar to what you wrote.

packages/n4s/src/rules/rules.js
```diff
@@ -82,5 +83,6 @@ export default function rules() {
     shorterThan,
     shorterThanOrEquals,
     startsWith,
+    isArrayOf,
```
ealush: Now that you added it to compounds, there's no need to have it here.

package.json
```diff
@@ -6,7 +6,7 @@
     "prettier-watch": "onchange '**/*.js' '**/*.json' -- prettier --write {{changed}}",
     "dev": "onchange -i './packages/**/src/**/*.js' -- yarn genJSConfig",
     "genJSConfig": "node ./scripts/genJsconfig",
-    "test": "jest --projects ./packages/*",
+    "test": "jest --projects ./packages/n4s/src/rules",
```
ealush:
```suggestion
    "test": "jest --projects ./packages/*",
```
You don't need to change this every time you want to run focused tests. If you want to just run the test for the rules, you can 
```yarn test --watch rules```

or ```yarn test --watch n4s``` to run all the tests for n4s

ealush (approved):
Nice! This looks amazing!

Now, just fix those last comments, and also, I also see there are some lint errors: https://travis-ci.com/github/ealush/vest/builds/201473568

I am approving this PR, but I'll merge once you make those last changes.

Thank you 🙏🏻

ealush:
That's amazing @Moses3301! Than you so much for helping out improving vest! I really appreciate it. Awesome job!

Issue #492, opened 2020-11-11
add enforce.loose()


ealush:
enforce.shape() supports defining an object schema, but it requires an exact match of an object. If the object contains extra keys, it will throw an error.

enforce.loose should implement the same behaviour of enforce.shape, but also allow extra keys to be present in the data object.

See [Shape](https://github.com/ealush/vest/blob/latest/packages/n4s/src/enforce/compounds/shape.js) for reference.

kapalex:
What do you think of the following implementation ?

```
export function shape(obj, shapeObj, loose) {
  for (const key in shapeObj) {
    const current = shapeObj[key];
    const value = obj[key];

    if (
      !asArray(current).every(fn =>
        fn(fn.name === optional.name ? [obj, key] : value)
      )
    ) {
      return false;
    }
  }

  if (!loose) {
    for (const key in obj) {
      if (!shapeObj[key]) {
        return false;
      }
    }
  }

  return true;
}

export function loose(obj, shapeObj) {
  return shape(obj, shapeObj, true);
}
```

Would also allow using shape with an additional param to get the same result (might be more clear than loose)

ealush:
Yes, that's pretty much what I had in mind. I would argue though that instead of a boolean argument, it might be better an options object with a `loose` property. This will prevent breaking changes if we want to add more capabilities in the future.

ealush:
@kapalex If you still want to take this issue, you can go ahead. I am done implementing my breaking changes :)

kapalex:
Should get to it today or tomorrow 😄

ealush:
closed by #505

PR #505, opened 2020-11-16
implement loose enforcement and add options to shape #492

kapalex:
| Q                | A   |
| ---------------- | --- |
| New feature?     | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✔ |
| Related issues   |     |

Implemented as we've discussed in the issue

### DIFF
--- a/packages/n4s/src/enforce/compounds/shape.js
+++ b/packages/n4s/src/enforce/compounds/shape.js
@@ -3,8 +3,10 @@ import runLazyRules from 'runLazyRules';
 /**
  * @param {Object} obj  Data object that gets validated
  * @param {Object} shapeObj  Shape definition
+ * @param {Object} options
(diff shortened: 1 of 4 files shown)


packages/n4s/src/enforce/compounds/shape.js
```diff
@@ -23,11 +25,15 @@ export default function shape(obj, shapeObj) {
     }
   }
 
-  for (const key in obj) {
-    if (!shapeObj[key]) {
-      return false;
+  if (!(options || {}).loose) {
```
ealush, line 28: I would usually opt for a default parameter (assuming options is either an object or undefined), but for now - since this is only limited to this line, and the transpiled size _is_ smaller with your suggested style, I'll take it 👍 .

ealush (approved):
Awesome! Thank you @kapalex! That's exactly what I had in mind.

Awesome job
💥

PR #526, opened 2020-11-25
added: [N4S] oneOf compound

### LINKED ISSUE #513: add enforce.oneOf compound for mutually exclusive validations
Similar to [anyOf](https://github.com/ealush/vest/blob/latest/packages/n4s/src/enforce/compounds/anyOf.js), we need a mutual exclusion rule: `oneOf`.

In the following example `value` must either be shorter than five OR longer than 10, but never both.
```js
enforce(value).oneOf(
  enforce.longerThan(10),
  enforce.shorterThan(5),
);
```

```js
  enforce.shape({
    id: enforce.isNumber(),
    name: enforce.shape({
      first: enforce.isString(),
      last: enforce.isString(),
    }),
  })
);
const Group = enforce.template(
  enforce.shape({
    id: enforce.isNumber(),
    name: enforce.isString(),
  })
);

enforce(value).oneOf(
  User,
  Group
);
```

hpsharon:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✔ |
| Related issues   |  513    |

Add oneOf functionality to determine if exactly one rule matches


packages/n4s/src/enforce/compounds/oneOf.js
```diff
@@ -0,0 +1,9 @@
+import runLazyRules from 'runLazyRules';
+
+/**
+ * @param {*} value   Value to be test against rules
+ * @param {Function[]} rules    Rules to validate the value with
+ */
+export default function oneOf(value, ...rules) {
+  return (rules) && (rules.filter(ruleGroup => runLazyRules(ruleGroup, value)).length === 1);
```
ealush: Cleaver use of filter! Nice!
It could be a bit wasteful and keep running even after more than two tests passed. I would have preferred to short circuit immediately as we detect more than one successful test.
Maybe we could try something similar to [this](https://github.com/ealush/vest/blob/latest/packages/anyone/src/one.js)?

ealush: The first condition doesn't check anything here. Since `rules` is an arguments spread, it will always be an array.
```suggestion
  return rules.filter(ruleGroup => runLazyRules(ruleGroup, value)).length === 1;
```

packages/n4s/src/enforce/compounds/oneOf.js
```diff
@@ -0,0 +1,9 @@
...
+}
```
ealush: Not a problem with your code - but I discovered recently that transpiling to ES5 actually makes rest parameters actually pretty large in terms of bundle size.

I added a simple workaround that mimics this behavior, but reduces its size quite dramatically

What the following change does is basically wrapping the function with a higher order function that passes down all the arguments as `first` argument, and the rest of the arguments as an array.

```suggestion
import runLazyRules from 'runLazyRules';
import { withFirst } from 'withArgs';

/**
 * @param {*} value   Value to be test against rules
 * @param {Function[]} rules    Rules to validate the value with
 */
function oneOf(value, rules) {
  return (rules) && (rules.filter(ruleGroup => runLazyRules(ruleGroup, value)).length === 1);
}

export default withFirst(oneOf)
```

packages/n4s/docs/compound.md
```diff
@@ -129,3 +129,11 @@ enforce({ data: [1, 2, 3] }).shape({
   data: enforce.isArrayOf(enforce.isNumber()),
 });
 ```
+
+## enforce.oneOf()
+
+enforce.oneOf can be used to determine if EXACTLY one of the rules applies. It will run against rule in the array, and will only pass if exactly one rule applies.
```
ealush:
```suggestion
enforce.oneOf can be used to determine if _exactly_ one of the rules applies. It will run against rule in the array, and will only pass if exactly one rule applies.
```

packages/n4s/docs/compound.md
```diff
@@ -129,3 +129,11 @@ enforce({ data: [1, 2, 3] }).shape({
...
+
+```js
+enforce(1).oneOf(enforce.isString(), enforce.isNumber());
```
ealush: This example is syntactically correct, but could be a little confusing because just one rule actually applies here. I would extend the comment a little bit to reduce the ambiguity.
```suggestion
enforce(value).oneOf(
  enforce.isString(),
  enforce.isNumber(),
  enforce.longerThan(1)
);

/*
value = 1      -> ✅ (value is a number)
value = "1"    -> ✅ (value is string)
value = [1, 2] -> ✅ (value is longer than 1)
value = "12"   -> 🚨 (value is both a string and longer than 1)
*/
```

ealush (approved):
👏🏻
Very nice! @hpsharon, we're almost there. I think we can make a few tiny changes to improve the performance and bundle size.

Thank you! 🙏🏻

ealush (approved):
👏🏻
Very nice! @hpsharon, we're almost there. I think we can make a few tiny changes to improve the performance and bundle size.

Thank you! 🙏🏻

hpsharon:
Hi @ealush,
I pushed some more changes. LMK what you think.

packages/n4s/src/enforce/compounds/oneOf.js
```diff
@@ -0,0 +1,14 @@
+import one from 'one';
+import runLazyRules from 'runLazyRules';
+import { withFirst } from 'withArgs';
+
+/**
+ * @param {*} value   Value to be test against rules
+ * @param {Function[]} rules    Rules to validate the value with
+ */
+function oneOf(value, rules) {
+  return one.apply(null,
```
ealush: >It could be a bit wasteful and keep running even after more than two tests passed. I would have preferred to short circuit immediately as we detect more than one successful test.
Maybe we could try something similar to this?

I may have mislead you with my previous comment. This works, but it adds some unneeded internal call to the `run` module function which performs some extra checks that we do not use here.

What I meant is that you implement similar logic, but not require the file itself. If you find it easily extractable, you could extract it to some shared location, but that's not a priority.

ealush:
Well done @hpsharon! 👏🏻
Thank you, this is perfect!

PR #530, opened 2020-11-29
Add an allOf() rule 

### LINKED ISSUE #521: add enforce.allOf compound for AND style requirements
Similar to [anyOf](https://github.com/ealush/vest/blob/latest/packages/n4s/src/enforce/compounds/anyOf.js), we need an AND relationship rule: `allOf`.

In the following example `value` must both be a string and longer than 10.
```js
enforce(value).allOf(
  enforce.longerThan(10),
  enforce.isString,
);
```

```js
  enforce.loose({
    id: enforce.isNumber()
    name: enforce.shape({
      first: enforce.isString(),
      last: enforce.isString(),
    }),
  })
);

const DisabledAccount = enforce.template(
  enforce.loose({
    disabled: enforce.equals(true)
  })
)

enforce(value).allOf(
  User,
  DisabledAccount
);
```

# Developer guidelines

1. Create a new file called `allOf` right next to anyOf. It should be pretty much the same, implementing the rules.some logic with some other counter mechanism.

2. To actually have it used by enforce, add it to this [compounds export](https://github.com/ealush/vest/blob/latest/packages/n4s/src/enforce/compounds/compounds.js#L6)

3. Add a new test file [here](https://github.com/ealush/vest/tree/latest/packages/n4s/src/enforce/compounds/__tests__) and test different scenarios

4. Add suitable types [here](https://github.com/ealush/vest/blob/latest/packages/vest/src/typings/enforce.d.ts#L90) and [here](https://github.com/ealush/vest/blob/latest/packages/vest/src/typings/enforce.d.ts#L196) (basically duplicate anyOf).

5. If you feel like it, also add suitable docs for this feature [here](https://github.com/ealush/vest/blob/latest/packages/n4s/docs/compound.md)

Moses3301:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✖ |
| Related issues   |     |

Add an allOf rule which validates a value for a set of rules

Related to issue #521

### DIFF
--- a/packages/vest/docs/_sidebar.md
+++ b/packages/vest/docs/_sidebar.md
@@ -16,3 +16,4 @@
   - [Excluding or including tests](./exclusion)
   - [Using with node](./node)
   - [Utilities](./utilities)
+- [Migration Guides](./migration)
--- a/packages/vest/docs/_sidebar.md.bak
+++ b/packages/vest/docs/_sidebar.md.bak
@@ -11,3 +11,4 @@
   - [Excluding or including tests](./exclusion)
   - [Using with node](./node)
   - [Utilities](./utilities)
+- [Migration Guides](./migration)
--- a/packages/vest/docs/index.html
+++ b/packages/vest/docs/index.html
@@ -14,6 +14,7 @@
       rel="stylesheet"
       href="https://cdn.jsdelivr.net/npm/docsify-themeable@0/dist/css/theme-simple.css"
     />
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/prismjs@1.22.0/themes/prism-tomorrow.css">
   </head>
   <body>
     <div id="app"></div>
--- a/packages/vest/docs/migration.md
+++ b/packages/vest/docs/migration.md
@@ -1,0 +1,107 @@
+# Migration guides
+
+## V2 to V3
+
+Vest version 3 comes with many new features, yet with a reduced bundle size. To achieve this, some redundant interfaces were removed. All v2 capabilities still exist, but the way to use some of them changed.
+
+**Replaced interfaces**
+
+### Removed: vest.get()
+
+From now on, use suite.get() to get the latest validation result.
+
+#### v2
+
+```js
+const suite = vest.create('user_form', () => {
+  /*...*/
+});
+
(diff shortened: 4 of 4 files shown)


packages/n4s/docs/compound.md
```diff
@@ -3,8 +3,9 @@
 Alongside the list of rules that only accept data provided by the user, enforce also supports compound rules - these are rules that accept other rules as their arguments. These rules let you validate more complex scenarios with the ergonomics of enforce.
 
 - [enforce.anyOf() - either/or validations](#anyof)
+- [enforce.allOf() - all/and validations](#anyof)
 - [enforce.shape() - Object's shape matching](#shape)
-  - [enforce.optional() - nullable keys](#optional)
+- [enforce.optional() - nullable keys](#optional)
```
ealush: The indentation here was intentional. Optional can only be used within `shape`. Can you please re-indent it?

packages/n4s/docs/compound.md
```diff
@@ -16,6 +17,14 @@ Sometimes a value has more than one valid possibilities, `any` lets us validate
...
 // A valid value would either an array or a string.
 ```
+## enforce.allOf() - all/and validations :id=allOf 
+
+Sometimes we need to make sure that a value for a set of rules, `all` lets us validate that a value passes _all_ of the supplied rules.
+
+```js
+enforce(value).allOf(enforce.isString(), enforce.longerThen(5));
+// A valid is string and longer then 5.
+```
```
ealush: Maybe also add the second example from the issue itself?

`````markdown

```js
  enforce.loose({
    id: enforce.isNumber()
    name: enforce.shape({
      first: enforce.isString(),
      last: enforce.isString(),
    }),
  })
);

const DisabledAccount = enforce.template(
  enforce.loose({
    disabled: enforce.equals(true)
  })
)

enforce(value).allOf(
  User,
  DisabledAccount
);
```
`````

ealush (approved):
👏🏻 Well done @Moses3301!
Thank you again!
There are only these two comments on the documentation, other than that, it looks great!

PR #533, opened 2020-11-30
added: [N4S] allOf compound rule

### LINKED ISSUE #521: add enforce.allOf compound for AND style requirements

Moses3301:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✖ |
| Related issues   |     |

Add an allOf rule which validates a value for a set of rules

Related to issue #521


packages/n4s/src/enforce/compounds/__tests__/allOf.test.js
```diff
@@ -0,0 +1,65 @@
...
+      );
+    });
+    it('Should succeed when all of the rules applies', () => {
+      expect(allOf('test', enforce.isString(), enforce.longerThan(3))).toBe(
+        true
+      );
+    });
+    it('Should fail with no rules', () => {
+      expect(allOf(3)).toBe(true);
+    });
```
ealush, line 18:
```suggestion
    it('Should pass when no rules are provided', () => {
      expect(allOf(3)).toBe(true);
    });
```

ealush (approved):
Well done @Moses3301! 👏🏻
Thank you very much!

Issue #393, opened 2020-09-21
Add test.each for iterative tests over an array


ealush:
Implement test.each (described here: https://github.com/ealush/vest/issues/356#issuecomment-684780967).

Similar to [jest's each function](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout), but slightly different. It can be discussed if we should support the `printf` formatting, or find an alternative.

ealush:
```js
test.each(table)('Name', 'MSG', (...args) => {})
```

ealush:
```js
z`Some message (${(arg1) => arg1})`
```

omriLugasi:
@ealush 
Currently there are 3 options for string interpolation (error message).
1.
```js
 test.each([[1,2,3]])('fieldName', '{{ a }} is greater then 5', (a, b, c) => {
    enforce(a).greaterThan(5)
})
// Will print 1 is greater then 5
```
2. With string format (like https://jestjs.io/docs/en/api#testeachtablename-fn-timeout)
3. ```  z`Some message (${(arg1) => arg1})` ``` Actually I cant understand how you want to implement [your](https://github.com/ealush/vest/issues/393#issuecomment-696283653) suggestion?

omriLugasi:
Another options is to let the user choose to put static error message as string or to pass function that return the error message.

For example: 

With Static error message
```js
test.each([[1,2,3]])('fieldName', 'This is a static error message', (a, b, c) => {
    enforce(a).greaterThan(5)
})

```

With generic error message

```js

const generateErrorMessage = (a, b, c) => `${a} is greater then 5`

test.each([[1,2,3]])('fieldName', generateErrorMessage, (a, b, c) => {
    enforce(a).greaterThan(5)
})

```
@ealush let me know what you think of this approach :)

ealush:
Hey @omriLugasi, sorry, I missed your message.

Yes, this is very much possible - and technically, it is the same as using a tagged function:
![image](https://user-images.githubusercontent.com/11255103/94586200-9fd14800-0289-11eb-8e51-2a35297007f1.png)

The reason I am reluctant to just let the consumer specify a function is that it prevents us from actually knowing whether it is their test call, or their string generation function.

Why? Because message is optional, so when the consumer does this:
```js
test.each([1,2,3])('fieldName', someFn);
```
We have no real way of knowing whether the consumer intended for the second argument to be the test call, or the string generation function.  Using our own tagged function allows us to know which function is really called.

.

[...]

PR #541, opened 2020-12-02
feat(vest): test.each

kapalex:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         |  |
| New feature?     | ✔ |
| Breaking change? |  |
| Deprecations?    |  |
| Documentation?   |  |
| Tests added?     | ✔ |
| Types added?     | ✔ |
| Related issues   |  #393  |

Seems to be working as intended, though your feedback would be appreciated.
One issue I'm not entirely sure of - 
I went ahead with the method @omriLugasi suggested for the statement - it can either be a string or a function.
The way I implemented test.each means if the statement is a function it will be called regardless the test results as it is generated prior to being passed on to test. 
I was wondering if you think this is the expected behavior, and if now should we change test interface to support argument-less function as statement (which would allow me to pass it on to be rendered later).
Let me know your thoughts and any changes you think should be made 😄

### DIFF
--- a/packages/vest/src/core/test/__tests__/each.test.js
+++ b/packages/vest/src/core/test/__tests__/each.test.js
@@ -1,6 +1,7 @@
 import faker from 'faker';
-import enforce from 'n4s';
 
+import VestTest from 'VestTest';
+import enforce from 'enforce';
 import test from 'test';
 import vest from 'vest';
 
@@ -18,6 +19,8 @@ describe("Test Vest's `test.each` function", () => {
         expect(testObjects).toHaveLength(2);
         expect(testObjects[0].failed).toBe(false);
         expect(testObjects[1].failed).toBe(false);
+        expect(testObjects[0]).toBeInstanceOf(VestTest);
+        expect(testObjects[1]).toBeInstanceOf(VestTest);
       });
 
       it('Should mark failed tests as such', () => {
@@ -50,27 +53,50 @@ describe("Test Vest's `test.each` function", () => {
         const testFn1 = jest.fn();
         const testFn2 = jest.fn();
 
-        const validate = vest.create(faker.random.word(), () => {
+        const res = vest.create(faker.random.word(), () => {
           vest.only('test2');
           test.each([[5, 4, 10]])('test1', statementFn1, testFn1);
           test.each([[5, 4, 10]])('test2', faker.lorem.sentence(), testFn2);
-        });
-        validate();
+        })();
         expect(testFn1).not.toHaveBeenCalled();
+        expect(res.tests.test1).not.toBeDefined();
         expect(statementFn1).toHaveBeenCalled();
         expect(testFn2).toHaveBeenCalled();
+        expect(res.tests.test2).toBeDefined();
       })
 
       it('Should work correctly with 1d-array', () => {
-        vest.create(faker.random.word(), () => {
-          testObjects = test.each([1, 2, 3])(faker.random.word(), faker.lorem.sentence(), (a) => {
+        const res = vest.create(faker.random.word(), () => {
+          testObjects = test.each([2, 1, 3])(faker.random.word(), faker.lorem.sentence(), (a) => {
             enforce(a).greaterThanOrEquals(2);
           });
         })();
         expect(testObjects).toHaveLength(3);
-        expect(testObjects[0].failed).toBe(true);
-        expect(testObjects[1].failed).toBe(false);
+        expect(testObjects[0].failed).toBe(false);
+        /* Since fieldName is shared between all result, if one fails we expect all to fail */
+        expect(res.hasErrors(testObjects[0].fieldName)).toBe(true);
+        expect(testObjects[1].failed).toBe(true);
+        expect(res.hasErrors(testObjects[1].fieldName)).toBe(true);
         expect(testObjects[2].failed).toBe(false);
+        expect(res.hasErrors(testObjects[2].fieldName)).toBe(true);
+      });
+
+      it('Should work with fieldName function', () => {
+        const res = vest.create(faker.random.word(), () => {
+          testObjects = test.each([2, 1, 3])((a) => `field${a}`, faker.lorem.sentence(), (a) => {
+            enforce(a).greaterThanOrEquals(2);
+          });
+        })();
+        expect(testObjects).toHaveLength(3);
+        expect(testObjects[0].failed).toBe(false);
+        expect(testObjects[0].fieldName).toBe('field2');
+        expect(res.hasErrors(testObjects[0].fieldName)).toBe(false);
+        expect(testObjects[1].failed).toBe(true);
+        expect(testObjects[1].fieldName).toBe('field1');
+        expect(res.hasErrors(testObjects[1].fieldName)).toBe(true);
+        expect(testObjects[2].failed).toBe(false);
+        expect(testObjects[2].fieldName).toBe('field3');
+        expect(res.hasErrors(testObjects[2].fieldName)).toBe(false);
       });
     });
   });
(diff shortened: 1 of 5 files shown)


kapalex:
Hey @ealush, just a quick reminder this is still up if you get the time to look it over

ealush:
Sorry @kapalex, I must have missed this PR. I'll go over it either today or tomorrow. 🙏🏻

packages/vest/src/core/test/test.each.js
```diff
@@ -0,0 +1,31 @@
...
+   * Run multiple tests using a parameter table
+   * @param {any[]} table         Array of arrays with params for each run (will accept 1d-array and treat every item as size one array)
+   * @param {String} fieldName    Name of the field to test.
+   * @param {String|function} [statement]  The message returned in case of a failure.  Follows printf syntax.
+   * @param {function} testFn     The actual test callback.
+   * @return {VestTest[]}           An array of VestTest instances.
+   */
+  function each(table) {
+    if (!Array.isArray(table)) {
+      throw new Error('[vest/test.each]: Expected table to be an array.');
```
ealush: Vest has a throwError utility. It already has the library name, and generally is a bit smaller.

```suggestion
      throwError('test.each: Expected table to be an array.');
```

packages/vest/src/core/test/test.each.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    return withFirst((fieldName, args) => {
+      const [testFn, statement] = args.reverse();
+
+      return table.map(item => {
+        item = asArray(item);
+        const itemStatement = typeof statement === 'function' ? statement(...item) : statement;
```
ealush: I just noticed that this is a pattern that I repeat quite often in the repo. created a new utility [optionalFunctionValue](https://github.com/ealush/vest/blob/latest/packages/__shared/src/optionalFunctionValue.js) function. You can use it instead.

packages/vest/src/core/test/test.each.js
```diff
@@ -0,0 +1,31 @@
...
+        return test(fieldName, itemStatement, () => testFn(...item));
```
ealush: Since this transpiles down to es5, I prefer to use apply to save a few bytes.
```suggestion
        return test(fieldName, itemStatement, () => testFn.apply(null, item));
```

packages/vest/src/core/test/test.each.js
```diff
@@ -0,0 +1,31 @@
...
+   * @param {function} testFn     The actual test callback.
+   * @return {VestTest[]}           An array of VestTest instances.
+   */
+  function each(table) {
+    if (!Array.isArray(table)) {
+      throw new Error('[vest/test.each]: Expected table to be an array.');
+    }
+
+    return withFirst((fieldName, args) => {
+      const [testFn, statement] = args.reverse();
```
ealush, line 22: Cool. I had a concern that it might not be suitable for this scenario, but now I see that it works well.

packages/vest/src/core/test/__tests__/each.test.js
```diff
@@ -0,0 +1,132 @@
+import faker from 'faker';
+import enforce from 'n4s';
```
ealush: Better use `enforce` directly in tests so you use it from source and not from bundle.

```suggestion
import enforce from 'enforce';
```

packages/vest/src/core/test/__tests__/each.test.js
```diff
@@ -0,0 +1,132 @@
...
+    describe('Basic Functionality', () => {
+      it('Should return all test objects correctly', () => {
+        vest.create(faker.random.word(), () => {
+          testObjects = test.each([[1, 2, 3], [2, 7, 9]])(faker.random.word(), faker.lorem.sentence(), (a, b, c) => {
+            enforce(a + b).equals(c);
+          });
+        })();
+        expect(testObjects).toHaveLength(2);
+        expect(testObjects[0].failed).toBe(false);
+        expect(testObjects[1].failed).toBe(false);
```
ealush, line 21: Since we're here, we can also make sure that it is an actual test object.
```suggestion
        expect(testObjects[1].failed).toBe(false);
        expect(testObjects[0]).toBeInstanceOf(VestTest);
        expect(testObjects[1]).toBeInstanceOf(VestTest);
        
```

packages/vest/src/core/test/__tests__/each.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+        const validate = vest.create(faker.random.word(), () => {
+          vest.only('test2');
+          test.each([[5, 4, 10]])('test1', statementFn1, testFn1);
+          test.each([[5, 4, 10]])('test2', faker.lorem.sentence(), testFn2);
+        });
+        validate();
+        expect(testFn1).not.toHaveBeenCalled();
+        expect(statementFn1).toHaveBeenCalled();
+        expect(testFn2).toHaveBeenCalled();
```
ealush: We can also test the field against the result:

```suggestion
        const result = validate();
        expect(testFn1).not.toHaveBeenCalled();
        expect(result.tests.test1).not.toBeDefined();
        expect(statementFn1).toHaveBeenCalled();
        expect(testFn2).toHaveBeenCalled();
        expect(result.tests.test2).toBeDefined();
```

packages/vest/src/core/test/__tests__/each.test.js
```diff
@@ -0,0 +1,132 @@
...
+      it('Should work correctly with 1d-array', () => {
+        vest.create(faker.random.word(), () => {
+          testObjects = test.each([1, 2, 3])(faker.random.word(), faker.lorem.sentence(), (a) => {
+            enforce(a).greaterThanOrEquals(2);
+          });
+        })();
+        expect(testObjects).toHaveLength(3);
+        expect(testObjects[0].failed).toBe(true);
+        expect(testObjects[1].failed).toBe(false);
+        expect(testObjects[2].failed).toBe(false);
```
ealush: Again, some integration could be useful here:
```suggestion
        const res = vest.create(faker.random.word(), () => {
          testObjects = test.each([1, 2, 3])(faker.random.word(), faker.lorem.sentence(), (a) => {
            enforce(a).greaterThanOrEquals(2);
          });
        })();
        expect(testObjects).toHaveLength(3);
        expect(testObjects[0].failed).toBe(true);
        expect(res.hasErrors(testObjects[0].fieldName)).toBe(true);
        expect(testObjects[1].failed).toBe(false);
        expect(res.hasErrors(testObjects[1].fieldName)).toBe(false);
        expect(testObjects[2].failed).toBe(false);
        expect(res.hasErrors(testObjects[2].fieldName)).toBe(false);
        
```

ealush (approved):
Very well done @kapalex, this is really great! There are some minor remarks but nothing serious. Sorry for taking such a long time to review this PR, I missed it.

packages/vest/src/core/test/test.each.js
ealush: Might be worth it to optionally allow passing a function as fieldName as well, so we can validate dynamically generated field names.

packages/vest/src/core/test/__tests__/each.test.js
kapalex: I applied this, and it actually revealed an interesting (and in my opinion correct) behavior
Since all test have the same field name - the result of hasErrors is always true
I've added another test case using the field name function I've added (which I agree was an important addition) to show the difference in behavior

kapalex:
Thanks for your comments @ealush, all applied and improvements made.
Let me know if there's anything else!

ealush:
This looks amazing @kapalex, thank you so much! This will be added to the new major version that will be released in a couple of weeks.

🙏🏻

PR #557, opened 2021-01-14
minor(vest): support schema skip and only

ealush:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✖ |
| Tests added?     | ✔ |
| Types added?     | ✔ |

Add a function body support to `vest/schema`. This allows using schema.skip/schema.only.

```js
import schema from 'vest/schema';
import {enforce} from 'vest'

const Name = enforce.shape({
  first: enforce.isString().isNotEmpty(),
  last: enforce.isString().isNotEmpty(),
  middle: enforc.optional(enforce.isString().isNotEmpty()),
});

const Schema = schema(Name, (data, ...someOtherArgs) => {
   schema.skip('middle'); // alternatively, you could use the passed args determine which field to skip
})

Schema({
  first: 'Jon',
  last: 'Bovi'
});
```

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -41,6 +41,7 @@
       "ruleMessage": ["./packages/n4s/src/meta/ruleMessage.js"],
       "ruleMeta": ["./packages/n4s/src/meta/ruleMeta.js"],
       "ruleWarn": ["./packages/n4s/src/meta/ruleWarn.js"],
+      "ruleWhen": ["./packages/n4s/src/meta/ruleWhen.js"],
       "rules": ["./packages/n4s/src/rules.js"],
       "endsWith": ["./packages/n4s/src/rules/endsWith.js"],
       "equals": ["./packages/n4s/src/rules/equals.js"],
--- a/packages/n4s/src/compounds/isArrayOf.js
+++ b/packages/n4s/src/compounds/isArrayOf.js
@@ -23,11 +23,15 @@ function isArrayOf(value, ruleChains) {
   }
 
   for (let i = 0; i < plainValue.length; i++) {
-    // Set result per each item in the array|
+    // Set result per each item in the array
     result.setChild(
       i,
       runCompoundChain(
-        EnforceContext.wrap(plainValue[i]).setFailFast(value.failFast),
+        new EnforceContext({
+          value: plainValue[i],
+          obj: plainValue,
+          key: i,
+        }).setFailFast(value.failFast),
         ruleChains,
         { mode: MODE_ANY }
       )
--- a/packages/n4s/src/meta/__tests__/__snapshots__/ruleMeta.test.js.snap
+++ b/packages/n4s/src/meta/__tests__/__snapshots__/ruleMeta.test.js.snap
@@ -1,5 +1,67 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Rule with "when" Should skip all excluded array items 1`] = `
+RuleResult {
+  "children": Object {
+    "items": RuleResult {
+      "children": Object {
+        "0": RuleResult {
+          "children": Object {
+            "checked": RuleResult {
+              "failed": false,
+              "hasErrors": false,
+              "hasWarnings": false,
+              "warn": false,
+            },
+          },
+          "failed": false,
+          "hasErrors": false,
+          "hasWarnings": false,
+          "when": false,
+        },
+        "3": RuleResult {
+          "children": Object {
+            "checked": RuleResult {
+              "failed": false,
+              "hasErrors": false,
+              "hasWarnings": false,
+              "warn": false,
+            },
+          },
+          "failed": false,
+          "hasErrors": false,
+          "hasWarnings": false,
+          "when": false,
+        },
+      },
+      "failed": false,
+      "hasErrors": false,
+      "hasWarnings": false,
+      "isArray": true,
+    },
+  },
+  "failed": false,
+  "hasErrors": false,
+  "hasWarnings": false,
+}
+`;
+
+exports[`Rule with "when" Should skip excluded object items 1`] = `
+RuleResult {
+  "children": Object {
+    "example1": RuleResult {
+      "failed": false,
+      "hasErrors": false,
+      "hasWarnings": false,
+      "warn": false,
+    },
+  },
+  "failed": false,
+  "hasErrors": false,
+  "hasWarnings": false,
+}
+`;
+
 exports[`Rule with a message Should add message to result 1`] = `
 RuleResult {
   "children": Object {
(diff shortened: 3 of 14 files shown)


therealpaulgg:
This looks pretty nice, I like the callback in the `schema` call. Allows for a lot of fine-grained control and would be perfect for my use case.

I'll put a sample code snippet of something similar I would be doing (the exact code I'm writing is unfortunately restricted to my work laptop). Will use typescript for some basic data structures.

```ts
import schema from "vest/schema"

import { enforce } from "vest"

interface FirstPhaseSchema {
    firstName: string
    lastName: string
    shippingDifferentFromBilling: boolean
    billingAddressLine: string
    shippingAddressLine: string
}

const firstPhaseSchema = enforce.shape({
    firstName: enforce.isString().isNotEmpty().message("First name is required"),
    lastName: enforce.isString().isNotEmpty().message("Last name is required"),
    shippingDifferentFromBilling: enforce.isBoolean(),
    billingAddressLine: enforce.isString().isNotEmpty(),
    shippingAddressLine: enforce.isString().custom((data: FirstPhaseSchema) => {
        // if checkbox is selected, shipping address must be filled out
        // otherwise, can just be an empty string value
        if (data.shippingDifferentFromBilling) {
            enforce.isString().isNotEmpty()
        } else {
            enforce.isString()
        }
    })
})

interface SecondPhaseSchema {
    creditCard: string
    cvc: number
}

const secondPhaseSchema = enforce.shape({
    creditCard: enforce.isString(),
    cvc: enforce.isNumber()
})

interface FormSchema {
    partOne: FirstPhaseSchema
    partTwo: SecondPhaseSchema
}

const formSchema = enforce.shape({
    partOne: firstPhaseSchema,
    partTwo: secondPhaseSchema
})

// vest schemas

const Schema = schema(formSchema, (data, fieldToSkip, ...someOtherArgs) => {
    schema.skip(fieldToSkip)
    // or is it schema.group?
})

const someData = {}

// validate first part
Schema(someData, "partOne")
// validate second part
Schema(someData, "partTwo")

// validate everything at once
Schema(someData)
```

The idea of this code is that I am abstracting away a large complex form structure into smaller, easier to manage form structures, but in such a way to where an entire complex form could be validated at the end to submit.

The calls like `.custom` is just me not knowing a better name for these calls. 

An 'odd' requirement I have is to know the number of fields that succeeded to those that failed (in order to show a progress bar for the form). So for example if the user had filled out first name, last name, and billing address, I would want to be able to see somehow that there were `n` number of tests completed and `k` number of errors, then I would be able to determine the percentage myself. This feature is already doable with `testsCount` I believe, but only at the outermost level.

ealush:
That's very interesting. First, that "custom" is already supported via [enforce.extend](https://ealush.com/vest/#/./n4s/custom?id=creating-custom-rules). I believe that's what you're aiming at.

I did also make another improvement to this branch, and added a "when" like feature (similar to the one you mentioned on discord).

Basically, it is another chained rule that determines whether a certain rule run should be skipped, either via a function call or a passed expression.

Here's an example:
```js
const data = {
  items: [
    { checked: true, id: '100' },
    { checked: false, id: '200' },
    { checked: false, id: '300' },
    { checked: true, id: '400' },
  ],
};

const Schema = enforce.shape({
  items: enforce.isArrayOf(
    enforce.shape({ checked: enforce.isBoolean(), id: enforce.isNumeric() })
	.when((value, key, parent) => {
      return value.checked; // we only validate checked items
    })
  ),
});

// This can also be used along with other enforce rules, since we have a boolean interface:
const Schema = enforce.shape({
  items: enforce.isArrayOf(
    enforce.shape({ checked: enforce.isBoolean(), id: enforce.isNumeric() })
	.when((value, key, parent) => {
      return enforce.greaterThan(200).test(value.id);
    })
  ),
});
```

The idea is that `when` takes a function. This function is given given the current rule value, the key (array index or the key in the object) and the parent object (either an array or an object). From what I tested it seems to work pretty well - you can take a look at the unit tests for a few more examples.

Combined with the aforementioned enforce.extened, would that be suitable for your scenario? If so, I'll publish a tagged version so you can test it out before the next release cycle.

therealpaulgg:
Yes, that looks great! With .when and .extend, seems like pretty much anything could be done. 

Not sure if `Schema(someData, "firstPart")` is possible or not but if that is as well then that's everything I would need for my use case. 

My only question would be: do you need to use `return` when using `enforce.greaterThan`?

Thanks for your work on this!

ealush:
If I understand correctly, by `Schema(someData, "firstPart")` you basically mean "skip firstPart".

If so, then yeah, it should work as expected.

```js
import schema from 'vest/schema';

const Schema = schema(formSchema, (data, fieldToSkip) => {
    schema.skip(fieldToSkip)
	// first time it would skip "partOne", second time it would skip "partTwo"
})

const someData = {}

// validate first part
Schema(someData, "partOne")
// validate second part
Schema(someData, "partTwo")
```

Alternatively, you also get schema.only that excludes all-other schema keys.

> My only question would be: do you need to use return when using enforce.greaterThan?

Yes. In the current implementation of `when` it needs to return with a boolean. A falsey value excludes the field run, and a truthy value includes it. I might add support for using enforce directly in there (something like `enforce(data).greaterThan(1)`) but I don't want to introduce too many changes all at once.

ealush:
@therealpaulgg I went ahead and published thi version: `vest@3.0.1-dev-32ceed`. It is exactly the same as the production version, only with the added features that I created in this PR. I would really appreciate it if you could report back on how it works for you and whether any changes should be made.

therealpaulgg:
Great, i'll test it as soon as i can. 

The intent of `Schema(someData, "partOne")` was to only validate `partOne`, it seems that `skip` would validate all other fields besides partOne (such as partTwo, three, four, etc). I'm looking for an easy way to validate __only__ `partOne`.

ealush:
> Great, i'll test it as soon as i can.
> 
> The intent of `Schema(someData, "partOne")` was to only validate `partOne`, it seems that `skip` would validate all other fields besides partOne (such as partTwo, three, four, etc). I'm looking for an easy way to validate **only** `partOne`.

Gotcha. Along with schema.skip I added schema.only, which filters out everything other than what you specified, so in our case:

```js
import schema from 'vest/schema';

const Schema = schema(formSchema, (data, fieldToInclude) => {
    schema.only(fieldToInclude)
	// first time it would include "partOne", second time it would include "partTwo"
})

const someData = {}

// validate first part
Schema(someData, "partOne")
// validate second part
Schema(someData, "partTwo")
```

ealush:
Included in https://github.com/ealush/vest/releases/tag/3.1.0 Documentation will follow once we leave experimental mode.

PR #599, opened 2021-03-02
Shani/branch fixed

shani-arnon:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✔/✖ |
| New feature?     | ✖ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✖ |
| Tests added?     | ✖ |
| Types added?     | ✖ |

- New branch includes two issues (added cache + expand suite)


packages/vest/src/core/produce/hasFailuresByGroup.js
```diff
@@ -1,4 +1,4 @@
-import { hasLogic } from 'hasFaillures';
+import { hasLogic } from 'hasFailures';
```
ealush, line 1: 🤦🏻‍♂️

packages/vest/src/core/produce/collectFailureMessages.js
```diff
@@ -11,35 +15,28 @@ import useTestObjects from 'useTestObjects';
...
+        collector[testObject.fieldName] = (
+          collector[testObject.fieldName] || []
+        ).concat(testObject.statement);
+
+        return collector;
+      }, {});
+
+      if (fieldName) return res[fieldName] || [];
+      return res;
+    })
```
ealush: Are you sure we need to bind the state to the context here? From the looks of it we do not use the state in this function or in any of the functions being called by it. I may be wrong about this - let's have a quick run of the tests without it to see if it actually works as expected.

```suggestion
    [testObjects, severity, options], () => {
      const res = testObjects.reduce((collector, testObject) => {
        if (group && testObject.groupName !== group) return collector;
        if (fieldName && testObject.fieldName !== fieldName) return collector;
        if (!testObject.failed) return collector;
        if (isMatchingSeverityProfile(severity, testObject)) return collector;

        collector[testObject.fieldName] = (
          collector[testObject.fieldName] || []
        ).concat(testObject.statement);

        return collector;
      }, {});

      if (fieldName) return res[fieldName] || [];
      return res;
    }
```

packages/vest/src/core/suite/createSuite.js
```diff
@@ -67,6 +67,14 @@ const createSuite = withArgs(args => {
...
     });
   });
+  suite.hasErrors = (fieldName) => suite.get().hasErrors(fieldName);
+  suite.hasWarnings = (fieldName) => suite.get().hasWarnings(fieldName);
+  suite.getErrors = (fieldName) => suite.get().getErrors(fieldName);
+  suite.getWarnings = (fieldName) => suite.get().getWarnings(fieldName);
+  suite.hasErrorsByGroup = (groupName, fieldName) => suite.get().hasErrorsByGroup(groupName, fieldName);
+  suite.hasWarningsByGroup = (groupName, fieldName) => suite.get().hasWarningsByGroup(groupName, fieldName);
+  suite.getErrorsByGroup = (groupName, fieldName) => suite.get().getErrorsByGroup(groupName, fieldName);
+  suite.getWarningsByGroup = (groupName, fieldName) => suite.get().getWarningsByGroup(groupName, fieldName);
```
ealush, line 82: Nice! This looks pretty good. Let's also add a couple of tests for these added functions. Nothing big, we only need to check that they match the original implementation.

Also, do you want to handle the documentation as well or should I?

packages/vest/src/typings/vest.d.ts
```diff
@@ -167,6 +167,27 @@ interface ICreateResult {
...
+  hasErrorsByGroup(groupName: string, fieldName: string): boolean;
+  
+  hasWarningsByGroup(groupName: string): boolean;
+  hasWarningsByGroup(groupName: string, fieldName: string): boolean;
+  
+  getErrorsByGroup(groupName: string): { [fieldName: string]: string[] };
+  getErrorsByGroup(groupName: string, fieldName: string): string[];
+  
+  getWarningsByGroup(groupName: string): { [fieldName: string]: string[] };
+  getWarningsByGroup(groupName: string, fieldName: string): string[];
```
ealush: Nice!
Here's an enhancement idea: Since all those are already typed here https://github.com/ealush/vest/blob/latest/packages/vest/src/typings/vestResult.d.ts#L28-L85 why not move them to their own type and use in both places? Would make maintenance way easier

ealush (commented):
Great job @shani-arnon! I left a couple of comments but overall this is looking pretty good.

Let's take care of these little details and get this one merged ✌️

Thank you for making Vest better 🙏🏻

packages/vest/src/core/produce/hasFailuresByGroup.js
shani-arnon: 🤓

packages/vest/src/core/produce/collectFailureMessages.js
shani-arnon: removed state bind. ran tests and works as expected

packages/vest/src/typings/vest.d.ts
shani-arnon: Done.

PR #612, opened 2021-04-26
feat(vest): add support for optional tests and "isValid"

ealush:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✔ |

# optional fields

> Since 3.2.0

It is possible to mark fields in your suite as optional fields. This means that when they are skipped, the suite may still be considered as valid.
All fields are by default required, unless explicitly marked as optional using the `optional` function.

## Usage

`optional` can take a field name as its argument, or an array of field names.

```js
import vest, { optional, only, test, enforce } from 'vest';

const suite = vest.create('RegisterPet', (data, currentField) => {
  only(currentField); // only validate this specified field

  optional(['pet_color', 'pet_age']);
  /** Equivalent to:
   * optional('pet_color')
   * optional('pet_age')
   **/

  test('pet_name', 'Pet Name is required', () => {
    enforce(data.name).isNotEmpty();
  });

  test('pet_color', 'If provided, pet color must be a string', () => {
    enforce(data.color).isString();
  });

  test('pet_age', 'If provided, pet age must be numeric', () => {
    enforce(data.age).isNumeric();
  });
});

suite({ name: 'Indie' }, /* -> only validate pet_name */ 'pet_name').isValid();
// ✅ Since pet_color and pet_age are optional, the suite may still be valid

suite({ age: 'Five' }, /* -> only validate pet_age */ 'pet_age').isValid();
// 🚨 When erroring, optional fields still make the suite invalid
```

## Difference between `optional` and `warn`

While on its surface, optional might seem similar to warn, they are quite different.
optional, like "only" and "skip" is set on the field level, which means that when set - all tests of an optional field are considered optional. Warn, on the other hand - is set on the test level, so the only tests affected are the tests that have the "warn" option applied within them.

Another distinction is that warning tests cannot set the suite to be invalid.

There may be rare occasions in which you have an optional and a warning only field, in which case, you may combine the two.

## `isValid` function

`isValid` returns whether the validation suite as a whole is valid or not.

A suite is considered valid if both conditions are met:

- There are no errors (`hasErrors() === false`) in the suite - warnings are not counted as errors.
- All non optional fields have passing tests.

```js
resultObject.isValid();
```


codechips:
Some thoughts.

**optional**

```js
// current implementation
optional(['pet_color', 'pet_age']);

// maybe nicer to make optional accept multiple args?
optional('pet_color', 'pet_age');
```

You can also flip the logic and make user state the required fields instead of optional. Makes the contract more explicit.

**isValid**

Not many validation frameworks get this functionality right or have it at all, because it's hard. I would like to eat the cake and also have it left. Let me explain.

When the user first loads the form we want to mark the form as invalid (disable the submit button) by running the validation suite. However, we don't want to show the validation errors upfront because the user hasn't yet interacted with the form.

Right now you have to do some weird gymnastics to achieve that in Vest. You have to validate the form on load and check `hasErrors` result that becomes your initial invalid state flag. After that you have to immediately reset the Vest state to prevent showing validation errors to the user.

Don't know how internals of the `isValid` work, but would be nice if it could be decoupled from the errors object. Maybe one way would be to pass a flag to it like `.isValid({ initial: true })` that validates the suite, but doesn't actually return errors or warnings in the result.

ealush:
@codechips, thank you for your quick and detailed reply 🙏🏻
It does confirm my thought that `isValid` would be a good addition to Vest. I hope that this answers your questions in term of the feature's usability.

**optional**

I tend to agree about `optional` accepting multiple params, I wanted to make it function like `skip` and `only` that accept an array. Maybe I should just use this opportunity to make all three functions take either an array or any number of string arguments.

I, too, thought about having a "required" function instead of "optional". This is how most other form validation libraries go about it. After looking around use-cases, I thought it would be best to go with the optional route for two reasons:

1. It would change the default. An existing form will be automatically invalid by this upgrade, and it would take consumers make changes to their forms to utilize `isValid`.
2. My assumption is that a form field would usually be required  - not all forms, of course. Some forms with large number of fields may have multiple optional fields, but the majority of online forms are more similar to sign-in/sign-up, comment and purchase forms - these tend to have more required fields than optional ones.


**isValid**

I agree that this one can be tricky, that's why I've waited so long before adding it - and I am only adding it now, because I believe in most cases it can bring value. Some corner cases that involve async tests may still require some more manual intervention - but for the most part, I think `isValid` hits the right balance.

As you noted, we currently don't have a way of distinguishing a validation suite that never ran from a suite that did run. There are two things that handle it in our case .

1. a suite is not considered to be valid if it has no test runs 
https://github.com/ealush/vest/pull/612/files#diff-f9f4273d9220ca2d8f5bdd314a598efeaea21ac960ed668e89f28dc69bd95d68R116-R118

2. a follow up PR that I am working on will initialize the suite with a default and empty test object that also lists the tests - so the result object will look somewhat like this:
```js
{
	name: 'login_form',
	testCount: 0,
	errorCount: 0,
	warnCount: 0,
	tests: {
		username: {
		  testCount: 0,
		  errorCount: 0,
		  warnCount: 0,
		  errors: [],
		  warnings: [],
		},
		password: {
		  testCount: 0,
		  errorCount: 0,
		  warnCount: 0,
		  errors: [],
		  warnings: [],
		}
	},
	groups: {},
	isValid: [Function],
	hasErrors: [Function],
	getErrors: [Function],
	hasWarnings: [Function],
	getWarnings: [Function],
    done: [Function],
	// ...
}
```
----


These two changes mean that you will able to use isValid right from the start, and in case this doesn't work for a specific edge case, you'll also have this initial empty object that you can write custom conditions for.

Issue #676, opened 2021-08-05
.isEmpty() doesn't check for whitespaces


sergiocasao:
`isNotEmpty` returns no errors when `data.username` is just whitespaces.

```
const suite = vest.create('NewUserForm', (data) => {

  test("username", "Username is required", () => {
    enforce(data.username).isNotEmpty();
  });

});

```

ealush:
Sorry @sergiocasao for replying late. Yes, this indeed seems like a feature that could be useful, but not as part of "isEmpty" as it would be a breaking change. Instead, @Pizzo15 added today the `isBlank` rule, which is available from `Vest@3.2.6`.

Again, sorry for taking a while - but this seems to be solving the issue.

Thanks for reporting this!

PR #698, opened 2021-10-08
feat(n4s): add isBlank rule

lpizzinidev:
Addresses #676

### DIFF
--- a/packages/n4s/docs/rules.md
+++ b/packages/n4s/docs/rules.md
@@ -30,6 +30,8 @@ Enforce rules are functions that allow you to test your data against different c
 - [isNotArray](#isnotarray)
 - [isBoolean](#isboolean)
 - [isNotBoolean](#isnotboolean)
+- [isBlank](#isblank)
+- [isNotBlank](#isnotblank)
 - [isNumber](#isnumber)
 - [isNotNumber](#isnotnumber)
 - [isNaN](#isNaN)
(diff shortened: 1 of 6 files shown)


ealush:
Hey @Pizzo15, thank you for this PR!

It makes great sense that Vest would come with this option, but this would be a major/breaking change and not a `patch` since it changes the way the library currently behaves.
Along with that, the current implementation is desired way, and I would much love to see your change appears in its own enforce rule, like so:

```
enforce("   ").isBlank(); // passes

enforce("   ").isNotBlank(); // throws
```

This should be easy enough to do.

Here are the general instructions:

1. Creating a new rule file:
Create a new file called "isBalnk.js" under [packages/n4s/src/rules](https://github.com/ealush/vest/tree/latest/packages/n4s/src/rules)

It should only cover strings, so the implementation should b quite simple as well.

2. Add a test file
Write the tests for the rule in the __tests__ directory.

3. Make enforce use the rule
Add the rule to the exported rules here: https://github.com/ealush/vest/blob/latest/packages/n4s/src/rules.js

4. Update the docs
Add usage example here: https://github.com/ealush/vest/blob/latest/packages/n4s/docs/rules.md


I am more than happy to help with any question you might have!

Thanks again

lpizzinidev:
Hi @ealush, I've just implemented the `isBlank/isNotBlank` rule.
Let me know if there are any problems or changes to be made.
Thank you!

ealush (approved):
Awesome! Thank you for this change, @Pizzo15! Will be released soon.

ealush:
Published in 3.2.6

PR #760, opened 2022-01-08
add(n4s): isValueOf and isNotValueOf

### LINKED ISSUE #758: Add rule `isValueOf`
Add a new rule that checks whether a value is a value in an object.

```js
enforce("Bravo").isValueOf({
	a: "Alpha",
	b: "Bravo",
  	c: "Charlie",
});
```

It should pass if the value passed to enforce is a value in the passed object.

To start:

1. Add the new rule file under https://github.com/ealush/vest/tree/latest/packages/n4s/src/rules
2. Add tests for that rule under https://github.com/ealush/vest/tree/latest/packages/n4s/src/rules/__tests__
3. Reference and export that rule here https://github.com/ealush/vest/blob/latest/packages/n4s/src/runtime/rules.ts

saard:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✖ |
| Related issues   |  #758   |

Added 2 assestions:
* isValueOf - checks if a string is a value of any key in an object.
* isNotValueOf - checks if a string is not a value of any key in an object.

### DIFF
--- a/packages/n4s/src/rules/__tests__/isValueOf.test.ts
+++ b/packages/n4s/src/rules/__tests__/isValueOf.test.ts
@@ -1,19 +1,36 @@
 import { enforce } from 'enforce';
+import { isValueOf, isNotValueOf } from 'isValueOf';
 
 const objectToTest = {
   a: 'Alpha',
   b: 'Bravo',
   c: 'Charlie',
-} as Object;
+};
 
-describe('isValueOf', () => {
-  it('Should pass as "Bravo" is the value of key "b"', () => {
-    enforce('Bravo').isValueOf(objectToTest);
+describe('isValueOf tests', () => {
+  describe('When "Bravo" is the value of key "b"', () => {
+    it('Should return true using enforce', () => {
+      enforce('Bravo').isValueOf(objectToTest);
+    });
+    it('Should return true', () => {
+      expect(isValueOf('Bravo', objectToTest)).toBe(true);
+    });
+  });
+  describe('When "Delta" is not a value of any key', () => {
+    expect(isValueOf('Delta', objectToTest)).toBe(false);
   });
 });
 
-describe('isNotValueOf', () => {
-  it('Should pass "Delta" is not a value of any key in the object', () => {
-    enforce('Delta').isNotValueOf(objectToTest);
+describe('isNotValueOf tests', () => {
+  describe('When "Delta" is not a value of any key', () => {
+    it('Should return true using enforce', () => {
+      enforce('Delta').isNotValueOf(objectToTest);
+    });
+    it('Should return true', () => {
+      expect(isNotValueOf('Delta', objectToTest)).toBe(true);
+    });
+  });
+  describe('When "Bravo" is the value of key "B"', () => {
+    expect(isValueOf('Delta', objectToTest)).toBe(false);
   });
 });
--- a/packages/n4s/src/rules/__tests__/isValueOf.test.ts
+++ b/packages/n4s/src/rules/__tests__/isValueOf.test.ts
@@ -1,36 +1,72 @@
 import { enforce } from 'enforce';
 import { isValueOf, isNotValueOf } from 'isValueOf';
 
-const objectToTest = {
-  a: 'Alpha',
-  b: 'Bravo',
-  c: 'Charlie',
+const testObject = {
+  a: 'Bravo',
+  b: false,
+  c: 42,
+};
+
+const testObject2 = {
+  d: {
+    greet: 'hello',
+  },
+  e: null,
+  f: undefined,
 };
 
 describe('isValueOf tests', () => {
-  describe('When "Bravo" is the value of key "b"', () => {
-    it('Should pass when used in enforce', () => {
-      enforce('Bravo').isValueOf(objectToTest);
+  describe('When the value exists in the object', () => {
+    it('Should return true using enforce', () => {
+      enforce('Bravo').isValueOf(testObject);
+      enforce(42).isValueOf(testObject);
+      enforce(false).isValueOf(testObject);
+      enforce(null).isValueOf(testObject2);
+      enforce(undefined).isValueOf(testObject2);
     });
+
     it('Should return true', () => {
-      expect(isValueOf('Bravo', objectToTest)).toBe(true);
+      expect(isValueOf('Bravo', testObject)).toBe(true);
+      expect(isValueOf(42, testObject)).toBe(true);
+      expect(isValueOf(false, testObject)).toBe(true);
+      expect(isValueOf(null, testObject2)).toBe(true);
+      expect(isValueOf(undefined, testObject2)).toBe(true);
     });
   });
-  describe('When "Delta" is not a value of any key', () => {
-    expect(isValueOf('Delta', objectToTest)).toBe(false);
+  describe('When the value does not exist in the object', () => {
+    it('Should return false', () => {
+      expect(isValueOf('Alpha', testObject)).toBe(false);
+      expect(isValueOf(1, testObject)).toBe(false);
+      expect(isValueOf(true, testObject)).toBe(false);
+      expect(isValueOf(null, testObject)).toBe(false);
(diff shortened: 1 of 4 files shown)


packages/n4s/src/rules/isValueOf.ts
```diff
@@ -0,0 +1,13 @@
+import bindNot from 'bindNot';
+import { isNotNullish } from 'isNullish';
+import { isString, isPlainObject } from 'lodash';
+
+export function isValueOf(value: string, objectToCheck: object): boolean {
+  return (
+    isNotNullish(objectToCheck) &&
+    isPlainObject(objectToCheck) &&
+    isString(value) &&
```
ealush: Value can be anything, not just a string.
```suggestion
```

packages/n4s/src/rules/isValueOf.ts
```diff
@@ -0,0 +1,13 @@
+import bindNot from 'bindNot';
+import { isNotNullish } from 'isNullish';
+import { isString, isPlainObject } from 'lodash';
```
ealush: Vest doesn't bring any third party dependencies, let's remove this and use whatever we have within Vest instead.
```suggestion
```

packages/n4s/src/rules/isValueOf.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function isValueOf(value: string, objectToCheck: object): boolean {
+  return (
+    isNotNullish(objectToCheck) &&
+    isPlainObject(objectToCheck) &&
```
ealush: There's no real reason we should only allow object literals. For all I care, we can also detect properties of all sort of objects.
```suggestion
```

packages/n4s/src/rules/isValueOf.ts
```diff
@@ -0,0 +1,13 @@
...
+    isString(value) &&
+    Object.values(objectToCheck).includes(value)
+  );
```
ealush: I think we could squeeze in a bit more perf by using a loop directly. This example is more verbose, but it should be more performant when dealing with larger datasets.

```suggestion

  if (isNullish(objectToCheck)) {
    return false;
  }

  for (const key in objectToCheck) {
    if (objectToCheck[key] === value) {
      return true;
    }
  }
  
  return false;
```

packages/n4s/src/rules/__tests__/isValueOf.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import { isValueOf, isNotValueOf } from 'isValueOf';
+
+const objectToTest = {
+  a: 'Alpha',
+  b: 'Bravo',
+  c: 'Charlie',
+};
+
+describe('isValueOf tests', () => {
+  describe('When "Bravo" is the value of key "b"', () => {
```
ealush: The test doesn't care about the specific value, only that it matches, so the test description should not refer to specific value. This is the same for the rest of the tests in this suite.


```suggestion
  describe('When the value exists in the object', () => {
```

packages/n4s/src/rules/__tests__/isValueOf.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    it('Should return true using enforce', () => {
+      enforce('Bravo').isValueOf(objectToTest);
+    });
```
ealush: enforce doesn't return true, it just doesn't throw. Let's change the description to something that's more in line with that. Something like:

```suggestion
    it('Should pass when used in enforce', () => {
      enforce('Bravo').isValueOf(objectToTest);
    });
```

packages/n4s/src/rules/__tests__/isValueOf.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    it('Should return true', () => {
+      expect(isValueOf('Bravo', objectToTest)).toBe(true);
+    });
+  });
+  describe('When "Delta" is not a value of any key', () => {
+    expect(isValueOf('Delta', objectToTest)).toBe(false);
+  });
```
ealush: You're missing the `it` statement. This means that there's no test run here. Same on line 34.

```suggestion
  describe('When "Delta" is not a value of any key', () => {
    expect(isValueOf('Delta', objectToTest)).toBe(false);
  });
```

packages/n4s/src/rules/__tests__/isValueOf.test.ts
```diff
@@ -0,0 +1,36 @@
+import { enforce } from 'enforce';
+import { isValueOf, isNotValueOf } from 'isValueOf';
+
+const objectToTest = {
+  a: 'Alpha',
+  b: 'Bravo',
+  c: 'Charlie',
+};
+
```
ealush, line 17: Let's add a few more test cases - for example, when the passed object is not really an object (`null`, `undefined`, ...etc)

packages/n4s/src/rules/__tests__/isValueOf.test.ts
ealush: You can also add a negative case for when the value is not inside the object.

```js
expect(() => enforce('someValue').isValueOf({})).toThrow()
```

ealush (commented):
Nice job @saard. There are a few changes that I believe should be done, but overall, thank you for the quick response to the issue.

Let's handle these comments and get that merged.

packages/n4s/src/rules/isValueOf.ts
```diff
@@ -0,0 +1,13 @@
+import bindNot from 'bindNot';
+import { isNotNullish } from 'isNullish';
+import { isString, isPlainObject } from 'lodash';
+
+export function isValueOf(value: string, objectToCheck: object): boolean {
```
ealush: Just noticed the type here. Value should not be a string, but `any`, since anything can be an object value

saard: please take into account that if we accept any then the following use case will fail:
`
let welcome= {greeting:{ hebrew:'hi', english:'hello'}};
isValueOf({ hebrew:'hi', english:'hello'}, welcome);
`

ealush: Yes, that makes sense. `{ hebrew:'hi', english:'hello'}` is not a value in the welcome object. There's an identical value there, but it is not the same one. It is enough for this function to check for referential equality, and assume people refer to it correctly with a variable. 

And regardless, even in that case - what about other primitives? In that case it should have been `boolean|null|undefined|symbol|string|number`.

saard: I think we can leave out null and undefined since we are checking that it is not nullish and then take the rest.
anyways whatever we decide we can document

packages/n4s/src/rules/isValueOf.ts
saard: removed external dependencies

packages/n4s/src/rules/isValueOf.ts
saard: removed plain object check

packages/n4s/src/rules/__tests__/isValueOf.test.ts
saard: fixed wrong structured tests + added more tests

packages/n4s/src/rules/isValueOf.ts
saard: we set on checking referential equality for non-primitives

saard:
@ealush - Please review the changes and let me know if anything else is needed.
I don't know why the vercel is stuck at pending (or failed before) I don't have access.

ealush (approved):
Awesome job @saard, thank you for helping out here! Approved, and will be added to the next minor version of Vest. 💯

saard:
Thanks @ealush for the opportunity, I enjoyed it.

ealush:
Published in 4.0.2. Thanks!

Issue #757, opened 2022-01-07
Add new rule `isKeyOf`


ealush:
Add a new rule that checks whether a value is a key of an object.

```js
enforce("bananas").isKeyOf({
	apples: 5,
	bananas: 2,
  	cantelopes: 0,
});
```

It should pass if the value passed to enforce is a key in the passed object.

To start:

1. Add the new rule file under https://github.com/ealush/vest/tree/latest/packages/n4s/src/rules
2. Add tests for that rule under https://github.com/ealush/vest/tree/latest/packages/n4s/src/rules/__tests__
3. Reference and export that rule here https://github.com/ealush/vest/blob/latest/packages/n4s/src/runtime/rules.ts

galezra:
I'll take that :)

ealush:
Awesome. Let me know if you need anything from me

PR #771, opened 2022-01-09
feat(n4s): added 'isKeyOf' rule

galezra:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✔|
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✔ |
| Tests added?     | ✔ |
| Types added?     | ✖ |
| Related issues   |#757|

- isKeyOf: Checks whether a value is a key of an object
- isNotKeyOf: Checks whether a value is not a key of an object


packages/n4s/src/rules/isKeyOf.ts
```diff
@@ -0,0 +1,13 @@
+import bindNot from 'bindNot';
+import { isBoolean } from 'isBooleanValue';
+import { isNullish } from 'isNullish';
+import { isStringValue } from 'isStringValue';
+
+export function isKeyOf(key: string | symbol | number, obj: any): boolean {
+  const shouldTerminate =
+    isNullish(obj) || isBoolean(obj) || isStringValue(obj);
+
+  return shouldTerminate ? false : key in obj;
```
ealush: `shouldTerminate` doesn't cover all cases, for example - when the passed object is a `Symbol`.

I think it is enough to assume good intent, and follow the happy path here, thus not needing to validate the object. A few levels up the stack `enforce` actually does a try/catch, and errors thrown are treated as false.

I would simplify it like that:

```suggestion
export function isKeyOf(key: string | symbol | number, obj: any): boolean {
  return key in obj;
```

This might also cause some of the tests to fail, but then you just have to check that this function throws.

packages/n4s/docs/rules.md
```diff
@@ -1280,3 +1282,52 @@ enforce('10.12').isPositive(); //passes
 enforce(-10).isPositive(); // throws
 enforce('-10.12').isPositive(); // throws
 ```
+
+## isKeyOf
```
ealush: This is good, but just a request - I am in the process of replacing the docs folder and moving it someplace else. Can you please move this change to here: https://github.com/ealush/vest/blob/latest/website/docs/enforce/enforce_rules.md instead?

packages/n4s/src/rules/__tests__/isKeyOf.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  it('When the key exists in the object (numeric)', () => {
+    expect(isKeyOf(1976, TOP_GROSSING_MOVIES)).toBe(true);
+  });
+
+  it.each([undefined, null, false, true, Object, [], '', Function.prototype])(
+    'Should return false for %s object',
+    v => {
+      expect(isKeyOf(DUMMY_KEY, v)).toBe(false);
+    }
+  );
```
ealush: When you make the change that I suggested, some of these will break. Then you'll need to check that isKeyof throws.

packages/n4s/src/rules/__tests__/isKeyOf.test.ts
```diff
@@ -0,0 +1,36 @@
+import { isKeyOf } from 'isKeyOf';
```
ealush: These tests are great, but can you also add tests for the function when used within enforce?

```js
it('Should pass when ...', () => {
  enforce(...).isKeyOf({...}) // passing values
})
```

```js
it('Should fail when ...', () => {
  expect(() => enforce(...).isKeyOf({...})).toThrow() // failing values
})
```

ealush (commented):
Awesome job, @galezra. Let's make these changes and get this merged 🙏🏻

packages/n4s/docs/rules.md
galezra: Sure no problem. But you gave me the same path I pushed the code on, do you mean another path?

website/docs/enforce/enforce_rules.md
```diff
@@ -57,8 +57,11 @@ Enforce rules are functions that allow you to test your data against different c
 - [doesNotStartWith](#doesnotstartwith)
 - [isNegative](#isnegative)
 - [isPositive](#ispositive)
+- [isKeyOf](#iskeyof)
+- # [isNotKeyOf](#isnotkeyof)
 - [isValueOf](#isvalueof)
 - [isNotValueOf](#isnotvalueof)
+  > > > > > > > origin/latest:website/docs/enforce/enforce_rules.md
```
ealush: Looks like you had a tiny merge conflict here, let's fix it and we're good to go
```suggestion
- [isKeyOf](#iskeyof)
- [isNotKeyOf](#isnotkeyof)
- [isValueOf](#isvalueof)
- [isNotValueOf](#isnotvalueof)
```

galezra:
Hi @ealush I pushed the changes, but for some reason, the file  convention (codeql-analysis.yml) has changed and I can't push the revert changes. Let me know if there are any other steps to take.

ealush:
> Hi @ealush I pushed the changes, but for some reason, the file convention (codeql-analysis.yml) has changed and I can't push the revert changes. Let me know if there are any other steps to take.

Yeah, don't worry about that. It's just prettier messing with us. This is fine.

ealush (approved):
Awesome job @galezra!
I took the liberty of reordering the links in the docs content, but other than that, this looks great. Thank you so much! 
💯

ealush:
Published in 4.0.2. Thanks!

PR #923, opened 2022-08-14
add(n4s): enforce().message() for inline assertion message

shani-arnon:
CONTRIBUTING.md

| Q                | A   |
| ---------------- | --- |
| Bug fix?        ✖ 
| New feature?     ✔
| Breaking change? ✖ 
| Deprecations?    ✖ 
| Documentation?   ✖ 
| Tests added?     ✔
| Types added?      ✔
| Related issues   |     |

Add `.message` function to enforceEager
What should `message` do?
If the eager enforce fails, it should throw the string itself (and not an error object)


packages/n4s/src/runtime/enforceEager.ts
```diff
@@ -7,9 +7,17 @@ import { getRule, RuleValue, Args, RuleBase, KBaseRules } from 'runtimeRules';
 import { transformResult } from 'transformResult';
 
 type IRules = n4s.IRules<Record<string, any>>;
+type TModifiers = {
+  message: (input: string) => enforceEagerReturn;
+};
 
-export default function enforceEager(value: RuleValue): IRules {
-  const target = {} as IRules;
+type enforceEagerReturn = IRules & TModifiers;
```
ealush:
```suggestion
type EnforceEagerReturn = IRules & TModifiers;
```

packages/n4s/src/runtime/enforceEager.ts
```diff
@@ -42,28 +51,42 @@ export default function enforceEager(value: RuleValue): IRules {
...
       // 2. Call the rule within the context, and pass over the arguments passed to it
       // 3. Transform the result to the correct output format
-      const transformedResult = ctx.run({ value }, () =>
-        transformResult(rule(value, ...args), ruleName, value, ...args)
-      );
+      const transformedResult = ctx.run({ value }, () => {
+        return transformResult(rule(value, ...args), ruleName, value, ...args);
+      });
+
+      function renderMessage() {
```
ealush: It's not really "render", more like, enforceMessage

shani-arnon: Changed to `enforceMessage `

packages/n4s/src/runtime/enforceEager.ts
```diff
@@ -42,28 +51,42 @@ export default function enforceEager(value: RuleValue): IRules {
...
+        const shouldUseCustomMessage = !isNullish(customMessage);
+        if (shouldUseCustomMessage) return customMessage;
+        if (isNullish(transformedResult.message)) {
+          return `enforce/${ruleName} failed with ${JSON.stringify(value)}`;
+        } else {
+          return StringObject(transformedResult.message);
+        }
```
ealush: The else is unneeded here
```suggestion
        }

        return StringObject(transformedResult.message);
```

packages/n4s/src/runtime/enforceEager.ts
```diff
@@ -7,9 +7,17 @@ import { getRule, RuleValue, Args, RuleBase, KBaseRules } from 'runtimeRules';
...
 type IRules = n4s.IRules<Record<string, any>>;
+type TModifiers = {
+  message: (input: string) => enforceEagerReturn;
+};
 
-export default function enforceEager(value: RuleValue): IRules {
-  const target = {} as IRules;
+type enforceEagerReturn = IRules & TModifiers;
+
+export default function enforceEager(value: RuleValue): enforceEagerReturn {
```
ealush:
```suggestion
// eslint-disable-next-line max-lines-per-function
export default function enforceEager(value: RuleValue): enforceEagerReturn {
```

ealush (approved):
Nice, @shani-arnon! This looks awesome. 💯
One last thing is that we need to add it to enforce's documentation, but this can be done in a follow-up PR.

Issue #808, opened 2022-03-06
[EASY] Add singular form error selectors


ealush:
Add singular form of the following:

getErrors() -> getError()
getWarnings() -> getWarning()

The idea is to return the first error/warning of a field, instead of an array.
Unlike getErrors/getWarnings that can either take 0 or 1 arguments, getError/getWarning must receive a field name.

Function signature:
```ts
function getError(fieldName: string): string;
function getWarning(fieldName: string): string;
```

General guidelines:

Add the functions here:
https://github.com/ealush/vest/blob/latest/packages/vest/src/core/suite/produce/produceSuiteResult.ts#L25-L27

Add to the API reference here:
https://github.com/ealush/vest/blob/latest/website/docs/api_reference.md

Add full documentation here:
https://github.com/ealush/vest/blob/latest/website/docs/writing_your_suite/result_object.md

Also, don't forget about tests.

shani-arnon:
Hi 👋🏼
I'd like to take this task

jhowlett-scottlogic:
@shani-arnon are you still working on this?

shani-arnon:
@jhowlett-scottlogic 👋🏼
Yes, sorry for the late response

jhowlett-scottlogic:
@shani-arnon  Sorry but i do not have the time to work on it now. I had a period of free time but i am now busy again with other commitments

quinnah:
Hi! Could I work on this?

ealush:
Added in V5

PR #925, opened 2022-08-14
initial

shani-arnon:
| Q                | A   |
| ---------------- | --- |
| Bug fix?         ✖ 
| New feature?      ✔
| Breaking change? ✖ 
| Deprecations?    ✖ 
| Documentation?   ✔
| Tests added?     |✔
| Types added?     |✔

Add singular form of the following: [https://github.com/ealush/vest/issues/808]

getErrors() -> getError()
getWarnings() -> getWarning()

The idea is to return the first error/warning of a field, instead of an array.
Unlike getErrors/getWarnings that can either take 0 or 1 arguments, getError/getWarning must receive a field name.

### DIFF
--- a/packages/vest/src/core/suite/produce/summarySelectors/suiteSelectors.ts
+++ b/packages/vest/src/core/suite/produce/summarySelectors/suiteSelectors.ts
@@ -91,7 +91,6 @@ export function suiteSelectors(summary: SuiteSummary): SuiteSelectors {
 
     const result: GetFailureResponse = {},
       warnings = getWarnings();
-    console.log({ getWarnings: warnings });
     if (!Object.keys(warnings).length) return '';
     Object.keys(warnings).forEach(
       (key: string) => (result[key] = warnings[key][0])
--- a/packages/vest/src/core/suite/produce/summarySelectors/suiteSelectors.ts
+++ b/packages/vest/src/core/suite/produce/summarySelectors/suiteSelectors.ts
@@ -3,7 +3,9 @@ import { isPositive } from 'vest-utils';
 import { Severity, SeverityCount } from 'Severity';
 import {
   FailureMessages,
+  FailureMessage,
   GetFailuresResponse,
+  GetFailureResponse,
   SuiteSummary,
   TestsContainer,
 } from 'SuiteSummaryTypes';
@@ -12,8 +14,10 @@ import { gatherFailures } from 'collectFailures';
 // eslint-disable-next-line max-lines-per-function, max-statements
 export function suiteSelectors(summary: SuiteSummary): SuiteSelectors {
   const selectors = {
+    getError,
     getErrors,
     getErrorsByGroup,
+    getWarning,
     getWarnings,
     getWarningsByGroup,
     hasErrors,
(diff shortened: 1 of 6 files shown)


packages/vest/src/core/suite/produce/summarySelectors/suiteSelectors.ts
```diff
@@ -80,12 +84,32 @@ export function suiteSelectors(summary: SuiteSummary): SuiteSelectors {
 
   // Responses
 
+  function getWarning(): FailureMessage;
+  function getWarning(fieldName: string): string;
+  function getWarning(fieldName?: string): GetFailureResponse {
```
ealush, line 89: I don't think there's need for an overload here (and same with getError).
fieldName should be mandatory, as without a fieldName, we don't really know which field should yield the "one" message.
```suggestion
  function getWarning(fieldName: string): string;
```

packages/vest/src/core/suite/produce/summarySelectors/suiteSelectors.ts
```diff
@@ -80,12 +84,32 @@ export function suiteSelectors(summary: SuiteSummary): SuiteSelectors {
...
+    if (fieldName) return getWarnings(fieldName)[0] || '';
+
+    const result: GetFailureResponse = {},
+      warnings = getWarnings();
+    console.log({ getWarnings: warnings });
```
ealush: 👀

packages/vest/src/core/suite/produce/summarySelectors/suiteSelectors.ts
```diff
@@ -80,12 +84,32 @@ export function suiteSelectors(summary: SuiteSummary): SuiteSelectors {
 
   // Responses
 
+  function getWarning(): FailureMessage;
+  function getWarning(fieldName: string): string;
+  function getWarning(fieldName?: string): GetFailureResponse {
+    if (fieldName) return getWarnings(fieldName)[0] || '';
```
ealush, line 90: It's rather simplistic, I believe this line can conclude the entire function.

getWarnings already runs all the safety checks, and it always returns an array (you can see here https://github.com/ealush/vest/blob/e8dbbc9da7b6f2b2c7c02c49406075889b2f2308/packages/vest/src/core/suite/produce/summarySelectors/collectFailures.ts#L23), and returning undefined in case there are no failures seems reasonable to me.
I've seen people check things like:

`if (typeof errors[0] === "string")`, which means that even an empty string is a problem here.

Thus:
```suggestion
    return getWarnings(fieldName)[0];
```

website/docs/api_reference.md
```diff
@@ -90,7 +90,9 @@ After running your suite, the results object is returned. It has the following f
 
 - [hasErrors](./writing_your_suite/result_object.md#haserrors-and-haswarnings) - Returns true if the suite or the provided field has errors.
 - [hasWarnings](./writing_your_suite/result_object.md#haserrors-and-haswarnings) - Returns true if the suite or the provided field has warnings.
+- [getError](./writing_your_suite/result_object.md#geterror-and-getwarning) - Returns a string error in the suite, or for a specific field.
```
ealush, line 93:
```suggestion
- [getError](./writing_your_suite/result_object.md#geterror-and-getwarning) - Returns the first error string of a given field.
```

website/docs/api_reference.md
```diff
@@ -90,7 +90,9 @@ After running your suite, the results object is returned. It has the following f
...
 - [getErrors](./writing_your_suite/result_object.md#geterrors-and-getwarnings) - Returns an object with errors in the suite, or an array of objects for a specific field.
+- [getWarning](./writing_your_suite/result_object.md#geterrors-and-getwarnings) - Returns a string with a warning in the suite, or for a specific field.
```
ealush, line 95:
```suggestion
- [getError](./writing_your_suite/result_object.md#geterror-and-getwarning) - Returns the first warning string of a given field.
```

website/docs/writing_your_suite/result_object.md
```diff
@@ -154,9 +154,31 @@ resultObject.hasWarningsByGroup('groupName');
...
+
+If there is no error for the field, the function defaults to an empty string.
+
+```js
+resultObject.getError('username');
+// ""
+
+resultObject.getWarning('username');
+// ""
+```
```
ealush, line 177:
```suggestion
If there is no error for the field, the function will return undefined.

```js
resultObject.getError('username');
// undefined

resultObject.getWarning('username');
// undefined
```
```

PR #961, opened 2022-10-21
docs: clearer server side function name

MentalGear:
Maybe we could change the function name to make clear what it does and where it should be used.

Alternatively, maybe vest could check internally for NODE_ENV and if it runs server-side, automatically transform to stateless so this security aspect is abstracted away and solved by default, which would be the most favorable approach.

| Q                | A   |
| ---------------- | --- |
| Bug fix?         | ✖ |
| New feature?     | ✖ |
| Breaking change? | ✖ |
| Deprecations?    | ✖ |
| Documentation?   | ✔ |
| Tests added?     | ✖ |
| Types added?     | ✖ |
| Related issues   |     |

### DIFF
--- a/website/docs/understanding_state.md
+++ b/website/docs/understanding_state.md
@@ -95,7 +95,7 @@ To do that, all you need to do is wrap your suite initialization with a wrapper 
 ```js
 import { create } from 'vest';
 
-function suite(data) {
+function serversideStatelessCheck(data) {
   return create(() => {
     test('username', 'username is required', () => {
       enforce(data.username).isNotEmpty();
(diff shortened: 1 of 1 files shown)


ealush (approved):
Hey @MentalGear, just got back from a 6 weeks of travel. Thank you for this PR.

In general, I agree it might be confusing, and ideally, this is something that Vest should do for you. I am looking into a separate stateless interface for Vest that will unlock many new capabilities.

It requires some changes to the internals, so would only probably be out in the next major.

Approving for new, thank you for taking such interest in improving Vest.

PR #1032, opened 2023-06-25
types(vest): fix `AsyncTest`, async test does not expect any returns

vonagam:
| Q                | A   |
| ---------------- | --- |
| Types added?     | ✔/✖ |

Current type of `AsyncTest` implies that resolving async test with a string or false means something (I assume fail), but it is not. Async test is not supposed to return anything, any return is disregarded.

For record, it is trivial to add support for booleans as return type, just by adding `if (result === false) VestTestMutator.fail(testObject);` in promise [done callback](https://github.com/ealush/vest/blob/eeb8bceb6dc25a6ef702043b862c7379968d8597/packages/vest/src/core/test/testLevelFlowControl/runTest.ts#L104). I think it makes sense and I can change the PR to do so, but for now this is simply about reflecting current state.

### DIFF
--- a/packages/vest/src/core/test/TestTypes.ts
+++ b/packages/vest/src/core/test/TestTypes.ts
@@ -3,7 +3,7 @@ import { Maybe } from 'vest-utils';
 import { TFieldName } from 'SuiteResultTypes';
 
 export type TestFn = () => TestResult;
-export type AsyncTest = Promise<Maybe<string | false>>;
+export type AsyncTest = Promise<void>;
 export type TestResult = Maybe<AsyncTest | boolean>;
 
 export type WithFieldName<F extends TFieldName = TFieldName> = {


ealush:
Hey @vonagam, thank you for this PR and working to improve Vest.

It is true that an async test disregards any resolution value, but on the other hand, when rejected with these values, an async test does accept false or a string. Please see these sections of the Vest documentation:

[Rejecting a Promise](https://vestjs.dev/docs/next/writing_tests/the_test_function#rejecting-a-promise-asynchronous-tests)
[Rejecting with a string](https://vestjs.dev/docs/next/writing_tests/failing_with_a_custom_message#rejecting-with-a-string)

Do these make sense, or does it still fail in your scenarios? If so, Let's see how we can resolve the issue.

Thanks again for wanting to improve Vest🙏

vonagam:
Yeah, I know, but `Promise<T>` is only about resolving with `T`. Rejecting or throwing can be done with anything and cannot be typed.

ealush:
Ohhh you are right. Now I see what you mean. Thank you for this 🙏

Merging

ealush (commented):
Initiated the release cycle of `next` version. Should be out in 10 minutes or so.

ealush:
@vonagam Out. `vest@5.0.0-next-20230625-b942`

PR #1070, opened 2023-09-08
types(vest): add more method overloads to `SuiteSelectors` type

vonagam:
| Q                | A   |
| ---------------- | --- |
| Types added?     | ✔/✖ |

Just small improvement for `getError`, `getWarning`, `getErrors`, `getWarnings`, `getErrorsByGroup` and `getWarningsByGroup`.

All three variants - without argument / with required argument / with optional optional - need to be provided to cover all possible cases.

If only one provided - an optional argument (like in `getError`/`getWarning`) - then a result is not as specific as it could be: `getError()` without an argument should not say that the result can be a string and `getError("field")` cannot include an object in its result.

If only two provided - without and with required (like in rest of listed funcs) - then typescript forbids to call a function with an undefined, so `getWarnings(variable)` is treated as wrong if `variable` can be undefined even though the code can handle it.

### DIFF
--- a/packages/vest/src/suiteResult/selectors/suiteSelectors.ts
+++ b/packages/vest/src/suiteResult/selectors/suiteSelectors.ts
@@ -140,6 +140,8 @@ export function suiteSelectors<F extends TFieldName, G extends TGroupName>(
     return getFailures(summary, Severity.WARNINGS, fieldName);
   }
 
+  function getWarning(): Maybe<SummaryFailure<F, G>>;
+  function getWarning(fieldName: F): Maybe<string>;
   function getWarning(fieldName?: F): GetSingularResponse<F, G> {
     return getFailure<F, G>(Severity.WARNINGS, summary, fieldName as F);
   }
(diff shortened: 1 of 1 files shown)


ealush:
Hm. let me go over it. I did not experience it, and I am worried this will change the behavior of the overloads. I'll run a few tests

vonagam:
[Here](https://www.typescriptlang.org/play?ssl=10&ssc=31&pln=1&pc=1#code/C4TwDgpgBAYgrgOygXigbwFBW1AFASgC4oBnYAJwEsEBzAbix1wA9iE4BbAIwnKKnbdeDHFAD0YvMwD8bTjz7EyVWlAA+A+cIwBfDBgA2EYFABmiFFEQATCKeoRrUAIYkXCEC7fwEDcwgI6cUllahoMf1wAZnwgiU0hcgjEaK8EhXUrBFt7BEdYqCA)  is a simple demonstration in typescript playground.

ealush (approved):
Thanks @vonagam! Ran a couple of tests and this looks good 👍

ealush:
Released in 5.0.4

PR #1074, opened 2023-09-16
types: remove `void` from `Maybe<T>` and update `SuiteSelectors`

vonagam:
| Q                | A   |
| ---------------- | --- |
| Types added?     | ✔/✖ |

In #1070 I've referenced `Maybe` in exported type. Did not know that `Maybe` is an internal helper type. Because of this I've met errors of such sort:
```
The inferred type of 'getError' cannot be named without a reference to '.pnpm/vest-utils@1.0.2/node_modules/vest-utils'. This is likely not portable. A type annotation is necessary.
```

So this PR removes `Maybe` usage from `SuiteSelectors`. 

But while doing it I've encountered resistance from typescript because `Maybe<T>` is currently defined as `T | undefined | void`. I think it is mistake to include `void` because it is a special thing - something like `unknown` but one that you should not touch at all, only discard. So I've changed `Maybe` to simple `T | undefined`.

### DIFF
--- a/packages/vest-utils/src/utilityTypes.ts
+++ b/packages/vest-utils/src/utilityTypes.ts
@@ -12,7 +12,7 @@ export type Nullish<T = void> = Nullable<T> | Maybe<T>;
 
 export type Nullable<T> = T | null;
 
-export type Maybe<T> = T | undefined | void;
+export type Maybe<T> = T | undefined;
 
 export type OneOrMoreOf<T> = T | T[];
 
--- a/packages/vest/src/core/test/TestTypes.ts
+++ b/packages/vest/src/core/test/TestTypes.ts
@@ -6,7 +6,7 @@ type TestFnPayload = { signal: AbortSignal };
 
 export type TestFn = (payload: TestFnPayload) => TestResult;
 export type AsyncTest = Promise<void>;
-export type TestResult = Maybe<AsyncTest | boolean>;
+export type TestResult = Maybe<AsyncTest | boolean> | void;
 
 export type WithFieldName<F extends TFieldName = TFieldName> = {
   fieldName: F;
--- a/packages/vest/src/suiteResult/SummaryFailure.ts
+++ b/packages/vest/src/suiteResult/SummaryFailure.ts
@@ -1,5 +1,3 @@
-import { Maybe } from 'vest-utils';
-
 import { TIsolateTest } from 'IsolateTest';
 import { TFieldName, TGroupName } from 'SuiteResultTypes';
 import { WithFieldName } from 'TestTypes';
@@ -10,8 +8,8 @@ export class SummaryFailure<F extends TFieldName, G extends TGroupName>
 {
   constructor(
     public fieldName: F,
-    public message: Maybe<string>,
-    public groupName: Maybe<G>
+    public message: string | undefined,
+    public groupName: G | undefined
   ) {}
 
   static fromTestObject<F extends TFieldName, G extends TGroupName>(
--- a/packages/vest/src/suiteResult/selectors/suiteSelectors.ts
+++ b/packages/vest/src/suiteResult/selectors/suiteSelectors.ts
@@ -149,7 +149,7 @@ export function suiteSelectors<F extends TFieldName, G extends TGroupName>(
 
   function getWarning(): Maybe<SummaryFailure<F, G>>;
   function getWarning(fieldName: F): Maybe<string>;
-  function getWarning(fieldName?: F): GetSingularResponse<F, G> {
+  function getWarning(fieldName?: F): Maybe<SummaryFailure<F, G> | string> {
     return getFailure<F, G>(Severity.WARNINGS, summary, fieldName as F);
   }
 
@@ -161,7 +161,7 @@ export function suiteSelectors<F extends TFieldName, G extends TGroupName>(
 
   function getError(): Maybe<SummaryFailure<F, G>>;
   function getError(fieldName: F): Maybe<string>;
-  function getError(fieldName?: F): GetSingularResponse<F, G> {
+  function getError(fieldName?: F): Maybe<SummaryFailure<F, G> | string> {
     return getFailure<F, G>(Severity.ERRORS, summary, fieldName as F);
   }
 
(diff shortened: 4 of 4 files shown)


ealush:
Hey, @vonagam, I think I missed this diff. Sorry. Thank you for making this PR

Here's the issue I was trying to resolve using `void` in Maybe:
image


Basically, TS treats undefined and void differently in function returns. Didn't dig into it, but removing void from `Maybe` makes it so that tests without a return value will fail the compiler.

An alternative would be to add `void` directly to `TestResult`, but that's pretty much the same thing, isn't it? Maybe there's a smarter way that I am not aware of.

ealush:
Also, I am not fully sure what would cause the issue you're seeing. Can you put a repro on codesandbox and I'll try to look at it? Wondering if it is a pnpm thing, vest thing, or something completely different.

Tried a quick repro on CodeSandbox and I didn't experiecne it, so interesting to see why it is behaving like that.

vonagam:
About repo: I have not used codesandbox much but this is part of an example that produces the error:
```ts
export class Wrapper {
  //...
  getError() { // <- mentioned error will be there
    return this.suiteResult.getError();
  }
}
```
Need to have `"declaration": true` in tsconfig as this one about reexporting types.

About no return: No return functions can be assigned to type of functions with undefined return without a problem - [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAGlC8UAUBKBA+KBXAdgEwgDMBLHCPAbgCgqBjAexwGdgoAPALlgWTXkwDeAX2pA). But void functions (so typed as void) cannot be assigned to anything besides void functions - [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAGlC8UAUBKBA+KBXAdgEwgDMBLHCPAbgCgBjAexwGdgoQAuZNeTANzuLwJOGKAG8AvtXpMWADw5xEIakA). So what you shown can be made work without an error but something prevents it. Edit: Intersting, `() => {}` can be assigned to `() => undefined` but not to `() => undefined | boolean`, feels like a bug in typescript.

vonagam:
Annoying, so in https://github.com/microsoft/TypeScript/issues/36288 ts got ability to accept `() => {}` as `() => undefined` but not as `() => undefined | T` which I find inconsistent... There is also https://github.com/microsoft/TypeScript/issues/36239 which will solve the issue.

As for now, removed modification to `Maybe` and just purged its usage from selectors types.

vonagam:
For record, seems like `TestResult` is the only time when `void` is needed in `Maybe`, so it is still possible to remove `void` from `Maybe` and just define `TestResult` as `Maybe<...> | void`.

vonagam:
Ok, two things.

First, published my integration of [vest for svelte](https://github.com/vonagam/svelte-vest). So now it should be possible to see the errors I am talking about. Let's take [this code](https://github.com/vonagam/svelte-vest/blob/782d7003645733ee2cb9224534cfafd0fab17e78/src/api/FormApi.ts#L378C3-L380C4):
```ts
getError(): Suite.Failure<V, A> | undefined {
  return this.summaryStore.value.getError() as any;
}
```

Ideally I would want not to specify types and let everything be inferred:
```ts
getError() {
  return this.summaryStore.value.getError();
}
```

But if I do that and run `pnpm tsc` I currently get:
> The inferred type of 'getError' cannot be named without a reference to '.pnpm/vest-utils@1.1.0/node_modules/vest-utils'. This is likely not portable. A type annotation is necessary.

So I am required to specify a type, but if I do it like this:
```ts
getError(): Suite.Failure<V, A> | undefined {
  return this.summaryStore.value.getError();
}
```

Then I get an error that says:
> Type 'void' is not assignable to type 'SummaryFailure<Field<A>, string> | undefined'.

That's the problem I am trying to solve.

Second, so I've reverted PR to almost initial state where I removed `void` from `Maybe` (as title says), but, as you pointed out, changed `TestResult` type to be `Maybe<> | void` to match that use case (that's the only place where Maybe is used as a public return type for that purpose).

Thoughts?

ealush:
This seems reasonable. I created Maybe as a very simple helper utility type to help me cope with having to type things repeatedly, but OTOH, maybe that's not the best option when exposing a publicly facing API, and the best approach is to be explicit to begin with.

Let me just test this locally real quick, and if this works, I'll merge, and release a `@next` version just so we can test it. If everything works then, we can release a patch version early next week.

ealush (approved):
Tested locally, all the tests pass and the build runs correctly too. Merging, and releasing a @next version. Release will finish in 10 minutes approximately

ealush:
Also, @vonagam, the vest-for-svelte repo looks great! Thank you so much for working on it
May I add it to the community integrations section of the website?

https://vestjs.dev/docs/community_resources/integrations

vonagam:
Thanks. Yes, of course, have nothing against the library being listed there.

ealush:
OK, @vonagam, 
`vest@next` can already be install. The exact tag is: `vest@5.2.3-next-20231215-ce83`.

I'll add the resource to the website tomorrow, I'm after a very long international flight, so it will take me just a little bit more time. Thank you for all the effort you're putting into Vest!

vonagam:
Ok, with that version I can remove `as any` from `getError`/`getWarning` usages. Which is good.

But seems like I still need to specify return type because `Maybe` [is used](https://github.com/ealush/vest/blob/4c3442928b23dcc1c5354e1655e7ee10c94496d0/packages/vest/src/suiteResult/SummaryFailure.ts#L13-L14) in `SummaryFailure` definition. If it is replaced there with simple `| undefined` then errors go away.

Can I add this change? (Checking because you made a release, don't want to mess things up.)

ealush:
@vonagam, yes, sure. Go ahead. The release goes through a separate branch, no issues there.

ealush:
@vonagam - OK, @next now is `vest@5.2.3-next-20231215-fecc`. Let's try this one.
I'll head out for today, but if that works for you, I'll get that merged and published soon.

vonagam:
Yep, works without a problem.

ealush:
@vonagam, released as vest@5.2.3

Issue #1073, opened 2023-09-14
Subscribe


vonagam:
I am writing a wrapper/helper/utility around Vest. Right now there is only one way to get notified when a suite updates - adding `done` callback on a run result.

But that's covers only running and only if it is you who invoked a suite so you have access to a run result and not basic result.

I would like to give users access to a suite so that they could do resets/removals/runs but then I would not be able react to or get notified about those state changes. So the only possible solution is to either allow access through some wrapper of sorts that mirrors (and ties myself to specific version of) vest suite api and calls notifier on every mutation or to simply not give any access to a suite.

There is an internal `VestBus` which already tracks what is happening with a suite. Is it possible to add method for subscription to suite state changes?

ealush:
That's actually something I've had in mind for a while.

I was working (then abandoned) on a vest devtools extension that relies on that, so if you look up 2 year old branches, you'll find reference to exactly that (built upon the vest bus).

I think I can add that, but it will be a few days, since I am currently at a conference abroad.

If you want to submit a PR, feel free to do so. I'll gladly review it.

What I had in mind is:

Suite.subscribe(cb)

Cb will receive as arguments:

- Event name
- Event payload

ealush:
Thinking of it again, the event payload may not even be needed in this case, and we can decide not to add it for now. It is enough to know about the event, and then you can pull whatever data you need directly from the suite. No need to expose the internals there.

ealush:
@vonagam, let me know if you want to take it. Otherwise, I'll implement it.

vonagam:
> the event payload may not even be needed in this case

Ye, for my use case it is not needed.

> let me know if you want to take it. Otherwise, I'll implement it.

Do it. (I usually contribute when I become impatient from waiting too long for a thing that I need but you are super fast with responses and have the full picture in mind so I consider it a better choice.)

ealush:
Hey @vonagam, created this PR. https://github.com/ealush/vest/pull/1080
Please take a look at it. If this looks like what you had in mind as well, I'll release it as part of 5.1.0.

ealush:
Shipped in 5.1.0. Closing.

PR #1080, opened 2023-09-25
Add suite.subscribe method

ealush:
| Q                | A   |
| ---------------- | --- |
| New feature?     | ✔|
| Documentation?   | ✔ |
| Tests added?     | ✔ |
| Related issues   |  #1073     |

## Subscribing to Suite State Changes

You can subscribe to changes in the suite state by calling `suite.subscribe(callback)`. The callback will be called whenever the suite state changes internally.

```js
suite.subscribe(() => {
  const result = suite.get();
  // ... Do something with the result
});
```

### Unsubscribing from Suite State Changes

The `subscribe` method returns a function that you can call to unsubscribe from the suite state changes:

```js
const unsubscribe = suite.subscribe();
```

### DIFF
--- a/packages/vest-utils/src/__tests__/bus.test.ts
+++ b/packages/vest-utils/src/__tests__/bus.test.ts
@@ -78,4 +78,50 @@ describe('bus', () => {
     expect(spy1).not.toHaveBeenCalled();
     expect(spy2).toHaveBeenCalled();
   });
+
+  describe('"ANY" wildcard (*)', () => {
+    it('Should run the wildcard handler on any event', () => {
+      const bus = createBus();
+      const spy1 = jest.fn();
+      const spy2 = jest.fn();
+      const spy3 = jest.fn();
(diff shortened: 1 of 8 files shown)


ealush:
@vonagam Please take a look at this, I believe this answers what we discussed

vonagam:
Looks right.

But need a way to unsubscribe:
```ts
function subscribe(cb: CB) {
  const {off} = VestBus.on('*', () => {
    cb();
  });
  
  return off;
}
```

Also, interesting that every `Vest.test` call triggers event bus, I thought that there was some optimizations to call only once for those sync tests. Not a problem, just a note for myself to add debounce for a subscribe callback provided.

ealush:
@vonagam added the unsubscribe function as the function's return.

And yeah. basically, internal communication in vest is done using events. This prevents different parts of the framework from knowing of one another, so they just report their current state and Vest reacts to these changes. Same with test results.

PR #1246, opened 2026-03-03
feat(vest): add schema filtering via focus/only with pick/omit

ealush:
## Summary

- Wire n4s `pick`/`omit` into vest's suite runner so `focus({ only, skip })` narrows schema validation to specified fields
- Refactor `applySchemaFocus` and `buildFocusedSchemaInstance` for cleaner n4s vendor detection and intersection handling
- Extract `bindSuiteResultMethods` helper for result construction
- Add integration test for intersected only/skip behavior

## Stack
- PR 1/3: n4s pick/omit schema modifiers
- **PR 2/3** (this PR)
- PR 3/3: getData resolver on suite result

Depends on #1245

## Test plan
- [x] Integration test: intersected only/skip fields parse correctly
- [x] Existing schema runtime tests pass

PR #1247, opened 2026-03-03
feat(vest): add getData resolver to suite result

ealush:
## Summary

- Add `getData()` method to suite results: returns validated schema parse output when no errors exist, otherwise falls back to raw input data
- Expose `getData` directly on the suite instance for convenience
- Add type definitions for `getData` on `SuiteResult` and `SuiteTypes`
- Update all snapshots to reflect the new `getData` function on results

## Stack
- PR 1/3: n4s pick/omit schema modifiers
- PR 2/3: vest schema filtering via focus/only
- **PR 3/3** (this PR)

Depends on #1246

## Test plan
- [x] Unit tests for `getData` (returns parsed output on valid, raw input on errors, undefined without schema)
- [x] Updated snapshots for all integration/unit test suites


ealush:
Recreating as a new PR to retrigger bots

PR #1257, opened 2026-03-13
feat: expose focus configurations on suite result run object

ealush:
Adds `run.focus` to the suite result object, allowing users to inspect the exact focus and exclusion parameters (`only`, `skip`, `onlyGroup`, `skipGroup`) for the current run.

### Changes
- ✨ Added `run.focus` property to `SuiteSummary`.
- 🛠️ Implemented `snapshotFocus` to safely capture normalized, frozen arrays of current focus modifiers.
- 🧊 Ensured deep immutability using `freezeAssign` and `Object.freeze`.
- 🔄 Verified that focus data does not bleed into subsequent runs or mutate previous result references.
- 📚 Updated `accessing_the_result.md` documentation.

### Verification
- Added comprehensive unit tests in `focus.test.ts` covering:
  - Property mirroring for object and method-chain configurations.
  - Isolation between consecutive runs.
  - Snapshot immutability.
- All tests passing (`yarn test`).

## Summary by CodeRabbit

* **New Features**
  * Run results now include a run.focus field showing applied focus modifiers (only/skip and group variants, including array forms); focus snapshots are immutable and do not persist across runs.

* **Tests**
  * Added tests validating focus propagation, group-array handling, deep reflection in results, and isolation/non-mutation across consecutive runs.

* **Documentation**
  * Docs updated with examples for accessing and interpreting run.focus.

### DIFF
--- a/packages/vest/src/suite/__tests__/focus.test.ts
+++ b/packages/vest/src/suite/__tests__/focus.test.ts
@@ -89,6 +89,37 @@ describe('suite.focus: only', () => {
         expect(res.run.focus?.skipGroup).toEqual(['groupE', 'groupF']);
       });
     });
+
+    describe('consecutive runs', () => {
+      it('should not persist focus data from previous runs', () => {
+        const suite = vest.create(() => {});
+
+        // First run with focus
+        suite.only('field_1').run();
+
+        // Second run without focus
+        const res2 = suite.run();
+
+        // Should be undefined because the second run had no focus modifiers
+        expect(res2.run.focus).toEqual({});
+      });
+
+      it('should not mutate previous run results with new focus data', () => {
+        const suite = vest.create(() => {});
+
+        // First run with focus on field_1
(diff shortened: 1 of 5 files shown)


ealush (commented):
Implemented snapshotFocus, which takes the arrays passed via .only()/.skip() or object configs and deeply freezes them so they cannot be mutated via the result payload. Also added explicit single-assertion tests for arrays.

ealush:
Added tests that verify focus data is not persisted across consecutive runs and that older suite results are not mutated by subsequent runs. 🚀

ealush:
Addressed the inconsistent empty-value handling in ! I've also refactored the function into smaller helpers to reduce cyclomatic complexity and satisfy ESLint rules. 🛠️

ealush:
All nitpicks addressed! Fixed the typo in  and updated the isolation test comment to accurately reflect the expected empty object. ✨

PR #1258, opened 2026-03-13
fix: skip executing omitWhen callback when ignored

ealush:
This PR modifies the `omitWhen` isolate to evaluate its conditional earlier and entirely skip executing the provided callback if the condition evaluates to true. 

This correctly resolves issues where fields within `omitWhen` loops would incorrectly register with a `testCount` of 0 rather than being excluded from the test suite entirely. As a result of this optimization, the unused `omitted` state has also been removed from `SuiteContext`.

Along with this fix, several new stateful async and sync test validation suits have been added covering testing states and ensuring execution context evaluation remains unbalkanized across reordered tests or dynamic omission phases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Omitted fields are now treated as invalid (validation state now reflects omission).

* **Tests**
  * Expanded and adjusted tests for omission behavior, including new stateful scenarios and parity between sync/async flows.

* **Documentation**
  * Updated omitWhen docs and examples to reflect conditional execution semantics.

* **Breaking Changes**
  * Removed previously exposed omission-related hooks/APIs; adjust integrations accordingly.

### DIFF
--- a/packages/vest/src/__tests__/integration.byGroup.test.ts
+++ b/packages/vest/src/__tests__/integration.byGroup.test.ts
@@ -397,7 +397,7 @@ describe('Integration: byGroup selectors', () => {
       const result = suite.run(true);
       expect(result.hasErrorsByGroup('omit_group', 'field_1')).toBe(false);
       expect(result.hasErrorsByGroup('omit_group', 'field_2')).toBe(true);
-      expect(result.isValidByGroup('omit_group', 'field_1')).toBe(true);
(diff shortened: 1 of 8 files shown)


ealush:
I have addressed the PR feedback:
1. Refactored the stateful tests in `omitWhen.test.ts` to use a helper function and reduce duplication.
2. Updated the documentation in `omitWhen.ts`, `website/docs`, and `website/versioned_docs` to reflect that the callback is now skipped when the condition is met.

PR #1268, opened 2026-03-24
fix(ci): fix release pipeline and sync package versions

ealush:
Includes both the fix for the release pipeline (missing commits/tags) and the version synchronizations to ensure no mismatches on the next release.

## Summary by CodeRabbit

* **Chores**
  * Bumped package versions across the ecosystem (patch releases).
  * Updated dependency ranges to newer internal utility versions.
  * Improved CI release workflow: safer push handling (suppresses “no changes” noise), token-based push authentication, and commit step now bypasses verification hooks for release commits.


ealush:
I have incorporated your feedback by capturing the `git push` output and failing the step if it returns an error other than 'Everything up-to-date'.

